PR #48, opened 2021-09-09
add spot/perp market without index but with oracle pk

mschneider:


program/tests/program_test/cookies.rs
```diff
@@ -443,31 +390,29 @@ impl SpotMarketCookie {
...
-                liquidation_fee,
-                optimal_util,
-                optimal_rate,
-                max_rate,
-            ).unwrap(),
-        ];
+        let instructions = [mango::instruction::add_spot_market(
+            &mango_program_id,
+            &mango_group_pk,
+            &oracle_pks[mint_index],
```
mschneider, line 396: here's the actual diff

i tried first to get `mango_group_cookie.mango_group.oracles[mint_index]` but the value was stale. if there was a way to refresh the group this would be basically a two-line change (also need to remove mint_index) rather than a whole puncher through all layers.

PR #53, opened 2021-09-13
InitMangoAccount: removed rent sysvar account

skrrb:
The rent sysvar account is not required for the mango account
initialization in the instructions; processors.rs does not use it.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -49,7 +49,7 @@ pub enum MangoInstruction {
 
     /// Deposit funds into mango account
     ///
-    /// Accounts expected by this instruction (8):
+    /// Accounts expected by this instruction (9):
     ///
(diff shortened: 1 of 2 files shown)


program/src/instruction.rs
```diff
@@ -1001,7 +999,6 @@ pub fn init_mango_account(
         AccountMeta::new_readonly(*mango_group_pk, false),
         AccountMeta::new(*mango_account_pk, false),
         AccountMeta::new_readonly(*owner_pk, true),
-        AccountMeta::new_readonly(solana_program::sysvar::rent::ID, false),
```
mschneider, line 1004: we will most likely not change this one, as it would be an api breaking change

mschneider:
please open a new PR, in case you still care to make these changes

PR #55, opened 2021-09-27
Added accounts to logs where necessary (mango_group_pk, liqee, liqor)…

nicholasgodfreyclarke:
…. Fixed malformed json. Added logging to cache_perp_markets. Made 'init_health above zero' an error'.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -694,9 +694,11 @@ impl Processor {
 
         msg!(
             "cache_prices details: {{ \
+            \"mango_group_pk\": \"{}\", \
             \"oracle_indexes\": {:?}, \
             \"oracle_prices\": {:?} \
         }}",
+            mango_group_ai.key,
             oracle_indexes,
             oracle_prices
         );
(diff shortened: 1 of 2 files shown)


program/src/processor.rs
```diff
@@ -2577,7 +2613,7 @@ impl Processor {
             if init_health > ZERO_I80F48 {
                 liqee_ma.being_liquidated = false;
                 msg!("Account init_health above zero.");
-                return Ok(());
+                return Err(throw_err!(MangoErrorCode::NotLiquidatable));
```
dafyddd, line 2616: This will revert changes and we need to keep the change to liqee_ma.being_liquidated

program/src/processor.rs
```diff
@@ -2892,7 +2934,7 @@ impl Processor {
...
```
dafyddd, line 2937: same as above

program/src/processor.rs
```diff
@@ -2982,8 +3024,21 @@ impl Processor {
 
         // TODO OPT make this more efficient
         msg!(
-            "liquidate_perp_market details: {{ \"market_index\": {}, \"base_transfer\": {}, \"quote_transfer\": {}, \"bankruptcy\": {} }}",
+            "liquidate_perp_market details: {{ \
```
dafyddd, line 3027: We have to make sure the compute used here is not too bad. @riordanp Can you test this out?

dafyddd:
I think you have access to create a branch inside mango-v3 repo. That would be preferable because then I can pull your branch and just make the changes I commented on if the changes are minor

nicholasgodfreyclarke:
I've reverted the code where I throw an error on "Account init_health above zero" now - would you still like me to make a branch?

program/src/processor.rs
riordanp: Just tried this on devnet, the extra logging costs 35670 compute

dafyddd: Additional 35k or 35k in total? As long as it clears with a lot of extra compute leftover I think we're safe

PR #60, opened 2021-10-05
Fix serialization of instructions to match MangoInstruction::unpack()

ckamm:
Previously, these enums were serialized into four bytes - deserialization expects one.

NOTE: Will add tests. Would be surprising if this was actually broken, maybe I'm just missing a setting?

@dafyddd

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -545,20 +545,39 @@ pub enum MangoInstruction {
     /// 1. `[writable]` perp_market_ai - PerpMarket
     /// 2. `[signer]` admin_ai - MangoGroup admin
     ChangePerpMarketParams {
+        #[serde(serialize_with = "serialize_option_fixed_width")]
         maint_leverage: Option<I80F48>,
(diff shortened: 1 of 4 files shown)


mschneider:
We sometimes serialize them with 1, sometimes with 4

image

ckamm:
Hmm. The problem is that currently `MangoInstruction::unpack(&my_instr.pack())` may return nonsense. I'd say the `pack()` is broken, since `unpack()` is part of the public interface. I'll see if I can fix that without adjusting the enum serialization width globally.

ckamm:
Ah, the layouts that use 4 bytes are for `serum_dex::matching::Side`, which this patch doesn't touch.

ckamm:
Tests are added. The patch also covers fixing ChangePerpMarketParams::pack() now.

dafyddd:
Wow thanks for taking the time to figure out serde custom serialization. I chose to just leave it not working on the rust side.

PR #79, opened 2021-11-20
fix(deposit instruction): incorrect expected accounts count

cogoo:
Nine accounts are listed as required as per the instruction


dafyddd:
Thanks. Fixed this in branch v3.2.7.

PR #80, opened 2021-11-24
Add `get_best_offer_for_amount`

acamill:
As I'm trying to figure out the best quantity and price for a given of money to spend, would this make sense?

I cannot do that on my side cause the bids/asks from the order book are private (intentional I assume). This is not the final code, could be improved, I went for some clarity

There is probably something better to use than my `Order` struct also, more like a wip where I seek feedback and to validate if this could be added or not

### DIFF
--- a/program/src/matching.rs
+++ b/program/src/matching.rs
@@ -641,6 +641,11 @@ impl BookSide {
     }
 }
 
+pub struct Order {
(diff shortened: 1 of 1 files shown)


acamill:
Reworking the code as it's nor working rn

PR #81, opened 2021-11-25
Feature/get best order for amount

acamill:
As I'm trying to figure out the best quantity and price for a given amount of money to spend, would this make sense?

I cannot do that on my side cause the bids/asks from the order book at private. This is not the final code, I just seek guidance on what I wish to accomplish, there might be a smarter/more efficient way to do this

### DIFF
--- a/program/src/matching.rs
+++ b/program/src/matching.rs
@@ -641,6 +641,11 @@ impl BookSide {
     }
 }
 
+pub struct Order {
+    quantity: i64,
+    price: i64,
+}
+
 pub struct Book<'a> {
     bids: RefMut<'a, BookSide>,
     asks: RefMut<'a, BookSide>,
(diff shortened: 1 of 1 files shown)


program/src/matching.rs
```diff
@@ -704,6 +709,65 @@ impl<'a> Book<'a> {
...
+
+        for order in book_side {
+            // This order total value in quote lots
+            let order_size = order.quantity.checked_mul(order.price()).unwrap();
+            // How much base_lot we can fill for this order size
+            let quantity_matched = {
+                if quote_lot_left_to_spend < order_size {
+                    // we can finish the operation by purchasing this order partially
+                    // find out how much quantity that is in base lots
+                    quote_lot_left_to_spend.checked_div(order.price()).unwrap()
```
mschneider, line 734: you should decide here if you want to round down or up

mschneider:
https://github.com/project-serum/serum-dex/blob/master/dex/src/critbit.rs#L825 this is a good example of how to write a unit test for this kind of code in rust

acamill:
> https://github.com/project-serum/serum-dex/blob/master/dex/src/critbit.rs#L825 this is a good example of how to write a unit test for this kind of code in rust

I'm not sure I can implement such testing in a reasonable amount of time, too much domain knowledge. 

If testing is mandatory I'll need some more pointer, if I can get in touch with someone who did implement similar testing? (I saw Daffy did some for market order testing in the integration ones, but it's commented)

program/src/matching.rs
acamill: Updated

acamill: Currently rounded down as we discussed, seems safer to have less

acamill:
And... I think I'll implement the counter part for doing the same in BASE

dafyddd:
I made bids and asks public in branch v3.2.7 which should be going out soon. Is that acceptable?

acamill:
Yes that's fine by me if you guys rather not have this code here. I'm cool
with both

acamill:
Will use that in my code for now, let me know if that's wanted at some point. Thanks for making it public

PR #82, opened 2021-11-28
prepare avax perp listing

mschneider:
allow listing perp before spot market for base tokens with arbitrary decimals


dafyddd:
Looks good. There's another check in add_spot_market() which checks that mint.decimals == 6. Can you make it check for mint.decimals == token_info.decimals?

PR #86, opened 2021-12-05
Removing unecessary rent account

zoheb-mzs:
For CPI compatibility


ckamm:
Looks good to me, since `processor::init_mango_account` only reads the first three accounts.

Please also update the doc comment for `InitMangoAccount` in instructions.rs.

dafyddd:
Fixed and deployed in v3.2.11

PR #87, opened 2021-12-08
implemented change_max_mango_accounts; governance UI still needs to b…

dafyddd:
…e implemented

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,9 @@
 # Mango Program Change Log
 
-## v3.2.8
-Deployed: | Slot: 
-1. Add Change
+## v3.2.9
+Deployed: Dec 8, 2021 at 22:29:47 UTC | Slot: 110,669,751
+1. Add ChangeMaxMangoAccounts
+2. Add some checked math in MangoAccount and matching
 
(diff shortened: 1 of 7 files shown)


changelog.md
```diff
@@ -1,5 +1,9 @@
 # Mango Program Change Log
 
+## v3.2.8
+Deployed: | Slot: 
+1. Add Change
```
ckamm, line 5: You typically update this only after the deployment, so it's ok to look like a placeholder for now?

program/src/state.rs
```diff
@@ -200,7 +200,9 @@ pub struct MangoGroup {
     pub msrm_vault: Pubkey,
     pub fees_vault: Pubkey,
 
-    pub padding: [u8; 32], // padding used for future expansions
+    pub max_mango_accounts: u32, // limits maximum number of MangoAccounts.v1 (closeable) accounts
```
microwavedcola1, line 203: could you elaborate on how this would actually be used? code looks ok, but I am lost why we need this

changelog.md
dafyddd: yeah

PR #91, opened 2021-12-15
V3.3.0

dafyddd:


program/src/processor.rs
```diff
@@ -206,6 +211,159 @@ impl Processor {
...
+        root_bank_cache.check_valid(&mango_group, now_ts)?;
+
+        let borrow_amount = mango_account
+            .get_native_borrow(&mango_cache.root_bank_cache[token_index], token_index)?;
+        let deposit_amount = mango_account
+            .get_native_deposit(&mango_cache.root_bank_cache[token_index], token_index)?;
+
+        // Amount must be dust aka < 1 native spl token
+        if borrow_amount > ZERO_I80F48 && borrow_amount < ONE_I80F48 {
+            checked_change_net(
```
dafyddd, line 329: use transfer_token_internal when possible. It's a minor point but it makes sure the transfer always works

dafyddd: @riordanp

dafyddd:
I think there's another vulnerability here. What if the attacker creates a bunch of v0 accounts and upgrades them to v1 to close and recoup SOL one by one?

program/src/processor.rs
```diff
@@ -4796,6 +5049,133 @@ impl Processor {
...
+            bids_ai,            // write
+            asks_ai,            // write
+        ] = accounts;
+
+        let mango_group = MangoGroup::load_checked(mango_group_ai, program_id)?;
+
+        let mut mango_account =
+            MangoAccount::load_mut_checked(mango_account_ai, program_id, mango_group_ai.key)?;
+        check!(!mango_account.is_bankrupt, MangoErrorCode::Bankrupt)?;
+        check!(owner_ai.is_signer, MangoErrorCode::Default)?;
```
ckamm, line 5147: `MangoErrorCode::SignerNecessary`

program/src/processor.rs
```diff
@@ -1223,6 +1381,52 @@ impl Processor {
...
+        let accounts = array_ref![accounts, 0, NUM_FIXED];
+
+        let [
+            mango_group_ai,     // read
+            mango_account_ai,   // write
+            owner_ai,           // write, signer
+            dex_prog_ai,        // read
+            open_orders_ai,     // write
+            spot_market_ai,     // read
+            signer_ai,          // read
```
ckamm, line 1396: I see no checks on open_orders_ai and signer_ai, is that correct?

ckamm (commented):
First pass check on accounts and their constraints, no logic

program/src/matching.rs
```diff
@@ -1428,40 +1354,69 @@ impl<'a> Book<'a> {
...
+            max_depth,
+            now_ts,
+            &mut limit,
+            asks_keys,
+        )
+    }
 
-        // First do bids
+    /// Internal
+    fn cancel_all_bids_with_size_incentives(
```
ckamm, line 1390: It sounds like `cancel_all_bids_with_size_incentives` and `cancel_all_asks_with_size_incentives` are nearly the same function. Could they be merged and take the side as an arg?

program/src/processor.rs
```diff
@@ -188,20 +189,177 @@ impl Processor {
...
             rent.is_exempt(mango_account_ai.lamports(), size_of::<MangoAccount>()),
-            MangoErrorCode::Default
+            MangoErrorCode::AccountNotRentExempt
         )?;
         check!(owner_ai.is_signer, MangoErrorCode::SignerNecessary)?;
 
-        let _mango_group = MangoGroup::load_checked(mango_group_ai, program_id)?;
+        let mango_group = MangoGroup::load_checked(mango_group_ai, program_id)?;
+        check!(
+            mango_group.num_mango_accounts <= mango_group.max_mango_accounts,
```
ckamm, line 198: After program upgrade, account creation will fail until SetMaxMangoAccounts was called (ah, maybe it has already been called?)

Why is this check even needed for the unclosable v0 accounts?

program/src/processor.rs
```diff
@@ -188,20 +189,177 @@ impl Processor {
...
+        check_eq!(&mango_account.advanced_orders_key, &Pubkey::default(), MangoErrorCode::Default)?;
+
+        // Transfer lamports to owner
+        program_transfer_lamports(mango_account_ai, owner_ai, mango_account_ai.lamports())?;
+
+        let mut mango_group = MangoGroup::load_mut_checked(mango_group_ai, program_id)?;
+        mango_group.num_mango_accounts -= 1;
+
+        // Prevent account being loaded by program
+        mango_account.meta_data.is_initialized = false;
```
ckamm, line 268: TBH, `is_initialized` going `false` is a bit scary. I think some code may rely on the fact that new accounts are zero-initialized when freshly created. Could we write 0 all over the account data or something, to avoid people quickly re-initing the closed account and somehow retaining some old data?

program/src/processor.rs
```diff
@@ -188,20 +189,177 @@ impl Processor {
...
+        // Check no perp positions or orders
+        for perp_account in mango_account.perp_accounts.iter() {
+            check_eq!(perp_account.base_position, 0, MangoErrorCode::InvalidAccountState)?;
+            check_eq!(
+                perp_account.quote_position,
+                ZERO_I80F48,
+                MangoErrorCode::InvalidAccountState
+            )?;
+
+            check!(perp_account.is_liquidatable(), MangoErrorCode::InvalidAccountState)?;
```
ckamm, line 246: Why does it need to be liquidatable?

Ah, this actually checks bids/asks/taker_quote/taker_base. Rename function?

program/src/processor.rs
```diff
@@ -188,20 +189,177 @@ impl Processor {
...
+            check_eq!(open_orders_key, &Pubkey::default(), MangoErrorCode::Default)?;
+        }
+        // Check advanced orders account closed
+        check_eq!(&mango_account.advanced_orders_key, &Pubkey::default(), MangoErrorCode::Default)?;
+
+        // Transfer lamports to owner
+        program_transfer_lamports(mango_account_ai, owner_ai, mango_account_ai.lamports())?;
+
+        let mut mango_group = MangoGroup::load_mut_checked(mango_group_ai, program_id)?;
+        mango_group.num_mango_accounts -= 1;
```
ckamm, line 265: prefer checked math to avoid num_mango_accounts wrapping for some reason

program/src/processor.rs
ckamm: maybe check that `mango_account.spot_open_orders[market_index] == open_orders_ai.key`, even if it's probably covered by the CPI

program/src/processor.rs
```diff
@@ -4734,7 +4984,9 @@ impl Processor {
         mango_account.mango_group = *mango_group_ai.key;
         mango_account.owner = *owner_ai.key;
         mango_account.order_market = [FREE_ORDER_SLOT; MAX_PERP_OPEN_ORDERS];
-        mango_account.meta_data = MetaData::new(DataType::MangoAccount, 0, true);
+        mango_account.meta_data = MetaData::new(DataType::MangoAccount, 1, true);
+
+        mango_group.num_mango_accounts += 1;
```
ckamm, line 4989: The `// TODO - test passing in MangoAccount that already has some data in it` seems unsure if `seed_and_create_pda()` will fail when called on an account that has already been initialized. I think it will, because `system_instruction::allocate()` will bail on already-allocated accounts.

Nevertheless, please double check that. If one could call `create_mango_account` multiple times on the same account, one could inflate num_mango_accounts easily.

If it's safe, remove the TODO comment in all places it exists.

program/src/processor.rs
```diff
@@ -4796,6 +5049,133 @@ impl Processor {
...
+        let mut mango_group = MangoGroup::load_mut_checked(mango_group_ai, program_id)?;
+        let mut mango_account =
+            MangoAccount::load_mut_checked(mango_account_ai, program_id, mango_group_ai.key)?;
+
+        check!(owner_ai.is_signer, MangoErrorCode::SignerNecessary)?;
+        check_eq!(&mango_account.owner, owner_ai.key, MangoErrorCode::InvalidOwner)?;
+        check_eq!(mango_account.meta_data.version, 0, MangoErrorCode::InvalidAccountState)?;
+        check!(!mango_account.not_upgradable, MangoErrorCode::InvalidAccountState)?;
+        check!(
+            mango_group.num_mango_accounts < mango_group.max_mango_accounts,
```
ckamm, line 5114: minor: sometimes this check is <, sometimes it's <=

ckamm (commented):
Everything but Dust-related

program/src/processor.rs
```diff
@@ -188,20 +189,177 @@ impl Processor {
...
+        let root_bank_cache = &mango_cache.root_bank_cache[token_index];
+        root_bank_cache.check_valid(&mango_group, now_ts)?;
+
+        let borrow_amount = mango_account
+            .get_native_borrow(&mango_cache.root_bank_cache[token_index], token_index)?;
+        let deposit_amount = mango_account
+            .get_native_deposit(&mango_cache.root_bank_cache[token_index], token_index)?;
+
+        // Amount must be dust aka < 1 native spl token
+        if borrow_amount > ZERO_I80F48 && borrow_amount < ONE_I80F48 {
```
ckamm, line 328: So 1 BTC is dust? :)

There's some value in having a fixed dust threshold -- but maybe safer to do the check in quote price?

Also, there's a DUST_THRESHOLD constant, that's used for liquidation, but it's way less than 1.0.

ckamm: Ah _native_ tokens!

ckamm: Seems hard to exploit if you can't give yourself slightly less than one native token of borrow reliably.

Maybe we could nevertheless subtract the borrow amount from the deposit amount if the deposit amount isn't fully dusted?

program/src/processor.rs
riordanp: Most of the account properties are already in their default state if the program gets to this point, I will make sure the remaining properties are reset too

program/src/processor.rs
riordanp: Plan is to set this with a governance vote before deploying v3.3

program/src/matching.rs
```diff
@@ -1301,122 +1301,48 @@ impl<'a> Book<'a> {
...
+    pub fn cancel_all_side_with_size_incentives(
+        &mut self,
+        mango_account: &mut MangoAccount,
+        perp_market: &mut PerpMarket,
+        market_index: usize,
+        side: Side,
+        mut limit: u8,
+    ) -> MangoResult {
+        // TODO - test different limits
+        let now_ts = Clock::get()?.unix_timestamp as u64;
```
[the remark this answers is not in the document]
dafyddd: Solana devs have told me Clock time is always increasing. I don't know with what strength they can guarantee that, but probably the safer thing to do is to go through all the times we get timestamp and error if it's negative. 

But if Clock time is not monotonically increasing, many things will break even if it doesn't go into negative territory.

program/src/matching.rs
dafyddd: They probably could, but then I'd have to switch up greater than or less than symbols everywhere based on side. Going to be a bit ugly either way.

program/src/processor.rs
dafyddd: Yes, I was also unsure, @riordanp could you create a test that does this?

program/src/processor.rs
```diff
@@ -4796,6 +5071,140 @@ impl Processor {
...
+            program_id,
+            system_prog_ai,
+            mango_account_ai,
+            mango_account_seeds,
+            &[],
+        )?;
+        let mut mango_account: RefMut<MangoAccount> = MangoAccount::load_mut(mango_account_ai)?;
+
+        mango_account.mango_group = *mango_group_ai.key;
+        mango_account.owner = *mango_group_ai.key;
```
dafyddd, line 5110: I think the owner should be admin. If we ever needed to do anything with this account, it'd probably be via the DAO

program/src/matching.rs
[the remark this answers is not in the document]
ckamm: Thanks. Will probably convert to u64 and panic on failure in voter-stake-registry

program/src/matching.rs
ckamm: Yeah, perfectly ok to not do in this commit, just a general observation that it'd be easy for bid/ask code to get subtly out of sync right now

program/src/processor.rs
dafyddd: It's not possible to have both deposits and borrows be positive. Program disallows that elsewhere and much relies on that.

But still, what someone could do is borrow some amount, generate dust interest, then deposit the amount borrowed back in, settle the dust borrow, and borrow that amount back in one tx. This would only work on SPL tokens where 1 native token has a substantial value exceeding the cost of SOL tx fee. In BTC case, this is true because 1e-6 BTC = $0.05. 

One solution is to require all OpenOrders to be closed and settle all dust at once. That way, you know the value of all the dust together must be positive because otherwise the account would be liquidated. But this soln doesn't work because then we'd require sending in up to 16 NodeBank pubkeys and 16 RootBank pubkeys and it wouldn't fit in a tx. 

But overall, the most you can exploit out of this is the value in the DustAccount and so it's probably not worth it for anyone to run a script to take that value. Worst case, DustAccount won't be able to take borrows anymore and people will have to settle their dust borrows by first buying 1 lot size on serum dex and withdrawing all the native amount to their token account.

program/src/processor.rs
```diff
@@ -1225,6 +1394,58 @@ impl Processor {
...
+            dex_prog_ai,
+            open_orders_ai,
+            signer_ai,
+            owner_ai,
+            spot_market_ai,
+            &[&signers_seeds],
+        )?;
+
+        mango_account.spot_open_orders[market_index] = Pubkey::default();
+        mango_account.in_margin_basket[market_index] = false;
```
dafyddd, line 1444: in_margin_basket should only be modified in state.rs -- I could make it a private var, but easier to just follow this rule for now. A bug that could happen here is if an OpenOrders is in_margin_basket, but it can actually be closed (we know this can happen with current protocol). Then invoke_close_open_orders will work and in_margin_basket will be set to false, but num_in_margin_basket will not equal sum(in_margin_basket) which is required.

program/src/instruction.rs
```diff
@@ -1476,6 +1487,25 @@ pub fn create_mango_account(
...
 
+pub fn set_delegate(
+    program_id: &Pubkey,
+    mango_group_pk: &Pubkey,
+    mango_account_pk: &Pubkey,
+    owner_pk: &Pubkey,
+    delegate_pk: &Pubkey,
+) -> Result<Instruction, ProgramError> {
+    let accounts = vec![
+        AccountMeta::new(*mango_group_pk, false),
```
ckamm, line 1498: the comments suggest `mango_group_pk` is `readonly`

program/src/processor.rs
```diff
@@ -3735,7 +3778,7 @@ impl Processor {
 
         let mut liqor_ma =
             MangoAccount::load_mut_checked(liqor_mango_account_ai, program_id, mango_group_ai.key)?;
-        check_eq!(liqor_ai.key, &liqor_ma.owner, MangoErrorCode::InvalidOwner)?;
+        check_eq!(liqor_ai.key, &liqor_ma.owner, MangoErrorCode::InvalidOwner)?; // todo
```
ckamm, line 3781: what does the todo mean? todo: consider allowing delegate?

program/src/processor.rs
```diff
@@ -4476,7 +4525,10 @@ impl Processor {
 
         let mut mango_account =
             MangoAccount::load_mut_checked(mango_account_ai, program_id, mango_group_ai.key)?;
-        check!(&mango_account.owner == owner_ai.key, MangoErrorCode::InvalidOwner)?;
+        check!(
+            &mango_account.owner == owner_ai.key || &mango_account.delegate == owner_ai.key,
+            MangoErrorCode::InvalidOwner
+        )?;
```
ckamm, line 4531: Sure about this one? withdraw_msrm sounds withdraw-y, should the delegate be allowed to do this?

program/src/instruction.rs
microwavedcola1: thanks, correct, fixed

program/src/processor.rs
microwavedcola1: thanks, correct, fixed

program/src/processor.rs
microwavedcola1: thanks, correct, fixed

program/src/processor.rs
riordanp: Tested this, calling create twice fails with this error the second time:
```
Allocate: account Address { address: ..., base: None } already in use
```

program/src/processor.rs
```diff
@@ -1415,6 +1415,86 @@ impl Processor {
...
+            system_prog_ai,
+            open_orders_ai,
+            open_orders_seeds,
+            &[],
+        )?;
+
+        {
+            let open_orders = load_open_orders(open_orders_ai)?;
+
+            // Make sure this open orders account has not been initialized already
```
ckamm, line 1473: I'm pretty sure this check is unnecessary as `seed_and_create_pda()` will fail if the account exists already. As far as I remember @dafyddd tested this, and the // TODO comments on other uses of `seed_and_create_pda()` can be removed

program/src/processor.rs
```diff
@@ -1415,6 +1415,86 @@ impl Processor {
...
+        check!(owner_ai.is_signer, MangoErrorCode::InvalidSignerKey)?;
+        check!(!mango_account.is_bankrupt, MangoErrorCode::Bankrupt)?;
+
+        let open_orders_seeds: &[&[u8]] =
+            &[&mango_account_ai.key.as_ref(), &market_index.to_le_bytes(), b"OpenOrders"];
+        seed_and_create_pda(
+            program_id,
+            owner_ai,
+            &Rent::get()?,
+            size_of::<serum_dex::state::OpenOrders>() + 12,
```
ckamm, line 1462: correct, but mysterious. serum accounts always start with "serum" and end with "padding".

program/src/processor.rs
```diff
@@ -1415,6 +1415,86 @@ impl Processor {
...
+        )?;
+        check!(owner_ai.is_signer, MangoErrorCode::InvalidSignerKey)?;
+        check!(!mango_account.is_bankrupt, MangoErrorCode::Bankrupt)?;
+
+        let open_orders_seeds: &[&[u8]] =
+            &[&mango_account_ai.key.as_ref(), &market_index.to_le_bytes(), b"OpenOrders"];
+        seed_and_create_pda(
+            program_id,
+            owner_ai,
+            &Rent::get()?,
```
ckamm, line 1461: Reading this now: I thought the `Rent` sysvar needs to be passed in? Anchor requires it, but seems unnecessary? That could simplify some calls in voter stake registry etc @microwavedcola1

ckamm:
On CreateSpotOpenOrders: Looks good to me, with the exception of missing a test and an instruction-creation function.

PR #94, opened 2021-12-17
document fee payment and event queue for place order

mschneider:


dafyddd:
Can you send this into v3.3.0 branch? If I merge into main it will mess up the line numbers for people debugging issues on mainnet

dafyddd:
ok added to v3.3.0

PR #95, opened 2021-12-17
Bump syn version to 1.0.82

chenwainuo:
Hey team, is it possible to bump syn version to 1.0.82

This is to make syn compatible with Anchor's syn version so mango can be easily added to any Anchor project's dependencies.

If not that's OK too, I will keep using my own fork

### DIFF
--- a/mango-macro/Cargo.toml
+++ b/mango-macro/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 proc-macro = true
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


dafyddd:
I think I hardcoded syn to 1.0.74 because increasing it broke something in the past. I'll try again

dafyddd:
Ok added to v3.3.0 and deployed to devnet; if all is well this will go to mainnet soon

PR #96, opened 2021-12-18
setting timeout for pyth conf filter to 60s

godmodegalactus:
, so that if the confidence of price is low for a long time we get latest price every minute


dafyddd:
Why? what if price is 3x the true value like what happened with the MSOL debacle?

godmodegalactus:
> Why? what if price is 3x the true value like what happened with the MSOL debacle?

The idea behind it to update oracle price if it has not been updated for a long time. But i agree that it should not be updated if the confidence is very very low.

PR #97, opened 2021-12-19
mc/add alternative authority which cant withdraw

microwavedcola1:
So basically this is a quick patch for supporting account delegation. Use case is to that funds can delegate their accounts to e.g. employees who can do everything except withdraw assets.

I decided to just reference this in the mango account. For future expansions we can use a pda which would slightly more code.

In addition to withdraw I kept these instructions unchanged, let me know your thoughts. 
1. close_advanced_orders
2. close_mango_account
3. close_spot_open_orders
4. deposit_msrm
5. init_mango_account
6. set_alternative_mango_account_authority
7. withdraw
8. withdraw_msrm

If there are no obvious issues, please double check all the instructions where I am not optionally comparing the alternative authority.


microwavedcola1:
Reopening, accidentally merged an empty PR, target branch should be unchanged

microwavedcola1:
Unrelated test_perp_trigger_orders_basic looks flaky, keeps failing
```

---- test_perp_trigger_orders_basic stdout ----
thread 'test_perp_trigger_orders_basic' panicked at 'deletion of unused is ok: IoError(Custom { kind: TimedOut, error: "Client dropped expired request." })', program/tests/test_perp_trigger_orders.rs:122:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

dafyddd:
Looks good so far. My main opposition to this is the verbose name. I think a simple `delegate` would be much easier to say and also is consistent with the naming of SPL token program. Also `delegate` or even `operator` better describes the function.

PR #98, opened 2021-12-20
Tests: Fix program log output

ckamm:
The log scope for these changed in solana 1.9


microwavedcola1 (approved):
tested locally, works as expected

PR #99, opened 2021-12-21
WIP: mc/flash loan

microwavedcola1:
Remaining must do tasks
- [ ]  add fees
- [ ] security review
- [ ] test/working-example for margin trade

Open questions
- [ ] flash loan - keep or remove in favor of margin trade
- [ ] margin trade - keep or remove in favor of flash loan
- [ ] Some common use cases for flash loans is collateral swapping and preventing self liquidation, wouldn't mango program be an ideal cpi program in this case? Would we then want to remove the check to not be able to cpi into mango program? This condition is necessary for daffy's margin trade instruction since it passes signer seeds but not technically necessary for the flash loan instruction since that explicitly does token transfer before doing the cpi
- [ ] should we strive for abi closer to spl-token-lending?

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -1346,6 +1346,7 @@ impl MangoInstruction {
             58 => MangoInstruction::SetDelegate,
             59 => {
                 let (amount, _cpi_data_vec_len, cpi_data_vec_arr) = array_refs![data, 8, 8;..;];
+
                 MangoInstruction::FlashLoan {
                     liquidity_amount: u64::from_le_bytes(*amount),
                     cpi_data: cpi_data_vec_arr.to_vec(),
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -912,6 +912,12 @@ pub enum MangoInstruction {
     FlashLoan {
         liquidity_amount: u64,
     },
+
+    MarginTrade {
+        num_open_orders: u8,
+        num_tokens_used: u8,
+        cpi_data: Vec<u8>,
+    },
 }
 
 impl MangoInstruction {
@@ -1340,6 +1346,15 @@ impl MangoInstruction {
             59 => {
                 let data = array_ref![data, 0, 8];
                 MangoInstruction::FlashLoan { liquidity_amount: u64::from_le_bytes(*data) }
+            }
+            60 => {
+                let (num_open_orders, num_tokens_used, cpi_data) = array_refs![data, 1, 1; ..;];
+
+                MangoInstruction::MarginTrade {
+                    num_open_orders: u8::from_le_bytes(*num_open_orders),
+                    num_tokens_used: u8::from_le_bytes(*num_tokens_used),
+                    cpi_data: cpi_data.to_vec(),
+                }
             }
             _ => {
                 return None;
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -892,11 +892,11 @@ pub enum MangoInstruction {
     ///   . `[]` token_prog_ai - Token program id.
     ///   . `[]` flash_loan_prog_ai - Flash loan receiver program id.
     ///             Must implement an instruction that has tag of 0 and a signature of `(amount: u64)`
-    ///             This instruction must return the amount to the source liquidity account.
+    ///             This instruction must return the amount to the source account.
     ///   .. `[any]` Additional accounts expected by the receiving program's `ReceiveFlashLoan` instruction.
     ///
     ///   The flash loan receiver program that is to be invoked should contain an instruction with
-    ///   tag `0` and accept the total amount (including fee) that needs to be returned back after
+    ///   tag `0` and accept the total amount that needs to be returned back after
     ///   its execution has completed.
     ///
     ///   Flash loan receiver should have an instruction with the following signature:
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -1350,6 +1350,7 @@ impl MangoInstruction {
             58 => MangoInstruction::SetDelegate,
             59 => {
                 let (amount, _cpi_data_vec_len, cpi_data_vec_arr) = array_refs![data, 8, 8;..;];
+                assert!(usize::from_le_bytes(*_cpi_data_vec_len) == cpi_data_vec_arr.len());
 
                 MangoInstruction::FlashLoan {
                     loan_amount: u64::from_le_bytes(*amount),
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -899,6 +899,10 @@ pub enum MangoInstruction {
     ///   tag `0` and accept the total amount that needs to be returned back after
     ///   its execution has completed.
     ///
+    /// Data expected:
+    ///  . loan_amount - size of flash loan
+    ///  . cpi_data - cpi instruction data
+    ///
     ///   Flash loan receiver should have an instruction with the following signature:
     ///
     ///   0. `[writable]` source_token_ai - Source vault (matching the destination from above).
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -911,6 +911,7 @@ pub enum MangoInstruction {
     ///   }
     FlashLoan {
         liquidity_amount: u64,
+        cpi_data: Vec<u8>,
     },
 
     MarginTrade {
@@ -1344,8 +1345,11 @@ impl MangoInstruction {
             }
             58 => MangoInstruction::SetDelegate,
             59 => {
-                let data = array_ref![data, 0, 8];
-                MangoInstruction::FlashLoan { liquidity_amount: u64::from_le_bytes(*data) }
+                let (amount, _cpi_data_vec_len, cpi_data_vec_arr) = array_refs![data, 8, 8;..;];
+                MangoInstruction::FlashLoan {
+                    liquidity_amount: u64::from_le_bytes(*amount),
+                    cpi_data: cpi_data_vec_arr.to_vec(),
+                }
             }
             60 => {
                 let (num_open_orders, num_tokens_used, cpi_data) = array_refs![data, 1, 1; ..;];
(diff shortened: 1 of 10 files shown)


program/src/instruction.rs
```diff
@@ -880,6 +880,38 @@ pub enum MangoInstruction {
...
+    /// Make a flash loan.
+    ///
+    /// Accounts expected: 4 + any optional
+    ///
+    ///   . `[]` mango_group_ai - MangoGroup
+    ///   . `[]` signer_ai - Source token account authority
+    ///   . `[writable]` source_token_ai - Source token account from which loan would be acquired
+    ///   . `[writable]` destination_token_ai - Destination token account
+    ///   . `[]` token_prog_ai - Token program id.
+    ///   . `[]` flash_loan_prog_ai - Flash loan receiver program id.
```
dafyddd, line 893: I think receiving program having to implement instruction with tag 0 is too restrictive. Why not pass in the ix data?

program/src/processor.rs
```diff
@@ -5313,6 +5313,91 @@ impl Processor {
...
+            flash_loan_instruction_account_infos.push(account_info.clone());
+        }
+
+        let signers_seeds = gen_signer_seeds(&mango_group.signer_nonce, mango_group_ai.key);
+        invoke_transfer(
+            token_prog_ai,
+            source_token_ai,
+            destination_token_ai,
+            signer_ai,
+            &[&signers_seeds],
```
ckamm, line 5371: If we sign like this and allow the program arbitrary writable accounts via `remaining_ais`, then what keeps the program from affecting other mango accounts guarded by this signature?

ckamm: Nevermind, since this is just the transfer and not the cpi call it's fine.

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+
+        check!(program_id != flash_loan_prog_ai.key, MangoErrorCode::InvalidProgramId)?;
+
+        let mango_group = MangoGroup::load_checked(mango_group_ai, program_id)?;
+        let source_token = Account::unpack(&source_token_ai.try_borrow_data()?)?;
+
+        let flash_loan_amount = if liquidity_amount == u64::MAX {
+            source_token.amount
+        } else {
+            min(source_token.amount, liquidity_amount)
```
ckamm, line 5547: as written, it could just always be `min()` without special casing for `u64::MAX`. But probably cases where liquidity_amount < source_token.amount should error?

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+            source_token_ai,          // write
+            destination_token_ai,     // write
+            token_prog_ai,            // read
+            flash_loan_prog_ai,       // read
+        ] = fixed_ais;
+
+        check!(program_id != flash_loan_prog_ai.key, MangoErrorCode::InvalidProgramId)?;
+
+        let mango_group = MangoGroup::load_checked(mango_group_ai, program_id)?;
+        let source_token = Account::unpack(&source_token_ai.try_borrow_data()?)?;
```
ckamm, line 5542: Should use `TokenAccount::load_checked` to verify that this really is a token account?

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+            signer_ai,
+            &[&signers_seeds],
+            flash_loan_amount,
+        )?;
+
+        solana_program::program::invoke(
+            &Instruction {
+                program_id: *flash_loan_prog_ai.key,
+                accounts: flash_loan_instruction_accounts,
+                data: cpi_data,
```
ckamm, line 5586: Above it's suggested that the first 8 bytes of argument to the instruction would be the amount that needs to be repaid. I don't see that here. Communicating it to the instruction is important for the liquidity_amount = u64::MAX case -- the callee has no idea about how much needs to be repaid otherwise (unless the dest account had a zero balance before, I guess)

ckamm: Similarly, the doc comment about starting with a zero byte seems no longer correct.

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+            },
+            &flash_loan_instruction_account_infos[..],
+        )?;
+
+        let source_balance_after = Account::unpack(&source_token_ai.try_borrow_data()?)?.amount;
+        check!(
+            // todo: fees
```
ckamm, line 5593: Yeah, fees seem important. Right now, mango doesn't gain anything from allowing people to flash loan - it just adds attack surface.

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+            source_token_ai,
+            destination_token_ai,
+            signer_ai,
+            &[&signers_seeds],
+            flash_loan_amount,
+        )?;
+
+        solana_program::program::invoke(
+            &Instruction {
+                program_id: *flash_loan_prog_ai.key,
```
ckamm, line 5584: How does the flash loan program have the authority to transfer funds away from dest_token_ai? It's either an account with a PDA or an extra signature is passed though the remaining_ais?

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+
+            check!(
+                root_bank.node_banks.contains(node_bank_ai.key),
+                MangoErrorCode::InvalidNodeBank
+            )?;
+            let node_bank = NodeBank::load_mut_checked(node_bank_ai, program_id)?;
+            check_eq!(&node_bank.vault, vault_ai.key, MangoErrorCode::InvalidVault)?;
+            // pre_amounts.push(Account::unpack(&vault_ai.try_borrow_data()?)?.amount);
+            let token_account = TokenAccount::load_checked(vault_ai)?;
+            token_account.check_mango_reqs(&mango_group)?;
```
ckamm, line 5383: I like these extra checks. Maybe even have a single function `TokenAccount::load_mango_vault_checked(...)` so we can't forget to call `check_mango_reqs()`? 

Should all the other uses of `Account::unpack()` be replaced? What makes this instance special?

program/src/state.rs
```diff
@@ -2383,3 +2416,23 @@ impl AdvancedOrders {
         Ok(state)
     }
 }
+
+#[derive(Copy, Clone, Pod, Loadable)]
+pub(crate) struct TokenAccount(pub(crate) Account);
+
+impl TokenAccount {
```
ckamm, line 2423: maybe add a `pub fn account()/data() -> &Account`, to avoid having to use `.0` externally?

ckamm (commented):
First pass, still need to go over `margin_trade` more.

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+        // Validate MangoCache
+        let now_ts = Clock::get()?.unix_timestamp as u64;
+        let active_assets = UserActiveAssets::new(&mango_group, &mango_account, extra_user_assets);
+        let mango_cache = MangoCache::load_checked(mango_cache_ai, program_id, &mango_group)?;
+        mango_cache.check_valid(&mango_group, &active_assets, now_ts)?;
+
+        /**
+        There are two types of `Account`: either owned by Mango program id or not.
+        We are not worried about `Account` owned by Mango program because we disallow CPI into Mango
+            Therefore, Mango owned accounts cannot be modified
```
ckamm, line 5400: Would it be possible to write a forwarder program that calls back into the mango program and that preserves the signatures?

program/src/processor.rs
```diff
@@ -5313,6 +5318,286 @@ impl Processor {
...
+            .map(|cpi_ai| {
+                if cpi_ai.is_writable {
+                    AccountMeta::new(*cpi_ai.key, cpi_ai.is_signer)
+                } else {
+                    AccountMeta::new_readonly(*cpi_ai.key, cpi_ai.is_signer)
+                }
+            })
+            .collect();
+        let mut cpi_account_infos: Vec<AccountInfo> = cpi_ais.iter().map(|ai| ai.clone()).collect();
+        cpi_account_infos.push(cpi_prog_ai.clone());
```
ckamm, line 5475: why's the program_ai added?

program/src/processor.rs
dafyddd: Solana says this type of reentrancy is not allowed. Something to be verified from Solana engineers and/or Neodyme

program/src/processor.rs
dafyddd: requirement of Solana runtime that you pass in that account during CPI. If you notice, we do the same for every other CPI

dafyddd:
@microwavedcola1 

1. I don't know what you mean by preventing self liquidation. I'm not sure if that's a legitimate use case in Mango. Either way, I think it's further dangerous to allow recursion and we'd have to be extra sure or put more guards around it before allowing it
2. ABI similar to token lending would be good if easy to do. I haven't looked at token, lending; maybe you can take a look?

program/src/processor.rs
dafyddd: https://docs.solana.com/developing/programming-model/calling-between-programs#reentrancy

ckamm: Great to see that this is guarded against in the runtime!

program/src/processor.rs
microwavedcola1: Hmm, tried that, ending up with 
```
[2021-12-29T09:44:42.499794000Z DEBUG solana_program_runtime::stable_log] Program log: panicked at 'from_bytes>SizeMismatch', /Users/mc/.cargo/registry/src/github.com-1ecc6299db9ec823/bytemuck-1.7.3/src/lib.rs:119:3
```

@dafyddd are you sure the zero copy code is correct?

program/src/processor.rs
microwavedcola1: yes its a progam owned account

program/src/processor.rs
dafyddd: I just added `TokenAccount` in this branch. All other instances of `Account::unpack()` should be replaced once we test this out and make sure it's equivalent to `TokenAccount::load_checked()`. I did not wrap the checks together because I wanted TokenAccount::load_checked to be more general, but I can't think of a case where doing those extra checks would be harmful actually.

PR #102, opened 2021-12-30
Fix test_perp_trigger_orders reliability

ckamm:
With this patch 5 out of 5 CI runs passed.


dafyddd:
Usually I don't want to merge into main unless it's a new mainnet deployment but I think this is fine because it only modifies test files.

PR #106, opened 2022-01-06
Soteria and Cargo Audit Automation

silas-x:
This PR covers two new features:

1. Cargo Audit: automated scanning of Cargo crates for vulnerable dependencies
2. Soteria: automated smart contract vulnerability scanning

Cargo audit:
New GitHub action along with the .cargo/audit.toml configuration file that provides arguments for cargo audit. The config file is handy for e.g. ignores and tracking that with change control.

Soteria:
Built out the existing CI job to include Soteria as well as a couple of other minor tweaks as the job was breaking. Soteria requires Solana and bpf to be installed, so it's a sensible place to run it at the end of other tests.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,10 +1,10 @@
 # All of the options which can be passed via CLI arguments can also be
 # permanently specified in this file.
 
-# RUSTSEC-2020-0071 and RUSTSEC-2020-0159 are low severity vulnerable upstream Solana crates. Ignored for now.
+# RUSTSEC-2020-0071, RUSTSEC-2020-0159 and RUSTSEC-2021-0124 are low severity vulnerable upstream Solana crates. Ignored for now.
 
 [advisories]
-ignore = ["RUSTSEC-2020-0071","RUSTSEC-2020-0159"] # advisory IDs to ignore e.g. ["RUSTSEC-2019-0001", ...]
+ignore = ["RUSTSEC-2020-0071","RUSTSEC-2020-0159", "RUSTSEC-2021-0124"] # advisory IDs to ignore e.g. ["RUSTSEC-2019-0001", ...]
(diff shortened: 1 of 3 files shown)


mschneider (approved):

mschneider:
check already found one vuln. that we should add to the ignore list RUSTSEC-2021-0124

mschneider:
soteria run results in a lot of output issues, but the CI passes anyways:
- are the found warnings about numprec explicitly ignored?
- are failures of this kind generally ignored?
- are all issues ignored rn?

how can we make sure we don't miss if a new contributor adds unchecked math that should be checked?

silas-x:
> audit check already found one vuln. that we should add to the ignore list RUSTSEC-2021-0124

Have reported the vuln upstream with Solana and will add to ignore list here for now.

silas-x:
> soteria run results in a lot of output issues, but the CI passes anyways:
> 
> * are the found warnings about numprec explicitly ignored?
> * are failures of this kind generally ignored?
> * are all issues ignored rn?
> 
> how can we make sure we don't miss when a new contributor adds unchecked math that should be checked?

Yes, it passes anything at the moment as there is no

> soteria run results in a lot of output issues, but the CI passes anyways:
> 
> * are the found warnings about numprec explicitly ignored?
> * are failures of this kind generally ignored?
> * are all issues ignored rn?
> 
> how can we make sure we don't miss when a new contributor adds unchecked math that should be checked?

Yes, the current implementation doesn't fail the job based on Soteria findings as there currently is no way to ignore false positives. Currently working on the following as an update:

1. Ability to ignore false positives / things that Soteria flags, but we will not fix
2. When that is working and documented, enable failing based on findings. The tool doesn't have severity ratings though, so will have to be binary

I think it still adds value 'as is', hence I wanted to get it out there.

PR #110, opened 2022-01-11
mc/change spot market params

microwavedcola1:
Implements a basic change for spot market params, doesnt do the multi rate setting, would prefer doing that separately, also have added a test

Signed-off-by: microwavedcola1 <microwavedcola@gmail.com>


microwavedcola1:
@ckamm reason for adding apt-get update to the ci workflow was when I was faced with below error

 https://github.com/blockworks-foundation/mango-v3/runs/4771654634?check_suite_focus=true#step:3:20

image

program/src/processor.rs
```diff
@@ -5313,6 +5313,84 @@ impl Processor {
...
+        // checks rootbank is part of the group
+        let _root_bank_index = mango_group
+            .find_root_bank_index(root_bank_ai.key)
+            .ok_or(throw_err!(MangoErrorCode::InvalidRootBank))?;
+
+        let mut root_bank = RootBank::load_mut_checked(&root_bank_ai, program_id)?;
+        let mut info = &mut mango_group.spot_markets[market_index];
+
+        // Unwrap params. Default to current state if Option is None
+        let (init_asset_weight, init_liab_weight) = if let Some(x) = init_leverage {
```
ckamm, line 5355: probably for a separate commit, but this sounds a lot like `Option::map_or()`

program/src/processor.rs
```diff
@@ -5313,6 +5313,84 @@ impl Processor {
...
+        // set the params on MangoGroup SpotMarketInfo
+        info.liquidation_fee = liquidation_fee;
+        info.maint_asset_weight = maint_asset_weight;
+        info.init_asset_weight = init_asset_weight;
+        info.maint_liab_weight = maint_liab_weight;
+        info.init_liab_weight = init_liab_weight;
+
+        check!(root_bank.meta_data.version != version, MangoErrorCode::InvalidParam)?;
+
+        root_bank.meta_data.version = version;
```
ckamm, line 5390: why does this check and set the `meta_data.version`? I expected that'd be used for account data layout upgrades?

program/src/processor.rs
```diff
@@ -5755,6 +5833,28 @@ impl Processor {
...
+                Self::change_spot_market_params(
+                    program_id,
+                    accounts,
+                    maint_leverage,
+                    init_leverage,
+                    liquidation_fee,
+                    optimal_util,
+                    optimal_rate,
+                    max_rate,
+                    version,
```
ckamm, line 5855: another thing for a cleanup/refactoring: passing these as arguments is error prone. It'd be nice if `ChangeSpotMarketParams` was a struct that could just be passed onwards to the executor function

microwavedcola1: I will move this to the refactoring ticket, noted.

program/src/processor.rs
microwavedcola1: I did a mistake while porting, fixed it now.

ckamm (approved):

ckamm:
clippy complains about the `map_or()`

PR #113, opened 2022-01-13
UXD Composability (cargo tests)

mschneider:


.github/workflows/ci-uxd.yml
```diff
@@ -0,0 +1,143 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
```
microwavedcola1, line 5: Would these tests only run on main? I would add a v.** wildcard, release are/will-be done from branches like e.g. v3.3.0

.github/workflows/ci-uxd.yml
```diff
@@ -0,0 +1,143 @@
...
+          echo "//npm.pkg.github.com/:_authToken=${{ env.NODE_AUTH_TOKEN }}" >> ~/.npmrc
+
+      - name: Yarn install + use local version of Mango-v3 +  run tests
+        run: |
+          cd ./uxd-program
+          sed -i.bak 's/^mango.*/mango={path="..\/..\/..\/program\/",features=["no-entrypoint"]}/' ./programs/uxd/Cargo.toml
+          yarn install
+          chmod +x ./scripts/reset_program_id.sh
+          ./scripts/reset_program_id.sh
+          npx @project-serum/anchor-cli@${{ env.anchor_version }} test --skip-local-validator
```
microwavedcola1, line 143: new blank line to make git happy

.github/workflows/ci-uxd.yml
microwavedcola1: ah my bad, it should be v**, just looked up

.github/workflows/ci-uxd.yml
```diff
@@ -0,0 +1,143 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  
+permissions:
+  contents: read
+  pull-requests: write
+  
+env:
+  solana_version: 1.8.11
```
microwavedcola1, line 17: mango refers to solana version, ^1.8.1, but this should be fine, as long it doesnt break

.github/workflows/ci-uxd.yml
```diff
@@ -0,0 +1,143 @@
...
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
+          solana airdrop 1 -u d
```
microwavedcola1, line 125: nitpick: can one technically send the sol back to the faucet? - be a good citizen and all

ckamm:
An early warning compatibility test like this sounds nice!

I'm a bit concerned about the security implications of checking out a foreign repo and running code from it in the github action. I think they have pretty wide permissions. Possibly there's a way to drop these permissions in the steps that interact with the uxd code? Maybe this is already done?

acamill:
> Possibly there's a way to drop these permissions in the steps that interact with the uxd code?

Could you give more info on this if you had a specific solution in mind? I've googled a bit on downgrading permission during a GH workflow but not sure I had the right keywords

> I'm a bit concerned about the

After going through https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
I can see that this is not the most ideal way of doing things.
Maximilian and Daffy are part of our org and have access to the repo, but that's still not ideal.
In the coming weeks (depending of workload and launch), I'll work on transferring the mango-logic back to mango.

I also want to make a "anchor_mango" rust dep, hosted on mango possibly, well not directly related to the tests. (Need to do some rewrite)

acamill:
I think the test simply fails on the 0.2 mint amount as it's small enough to have this discrepancy shows. Basically the execution price and the "test validation price" might not be the same? A bit out of idea on this one
image

In the CI the exec price was `148.479` (perp price)
image

In the CI the perp price fetched from mango by the test was ` 148.4945` (perp price)
image



https://explorer.solana.com/tx/4dptKDYrL43vJ9b5isqvSRWqrQy76gpyCfUeDt3KinqBen5kX9rMzWtMQ9nAWey8vDtc2bf98vLqGh1nRUEwVsG2?cluster=devnet

ckamm:
> > Possibly there's a way to drop these permissions in the steps that interact with the uxd code?
> 
> Could you give more info on this if you had a specific solution in mind? I've googled a bit on downgrading permission during a GH workflow but not sure I had the right keywords

I was thinking specifically of the job having the permissions of the github org and potentially being able to read private repos or make commits. But after reading https://docs.github.com/en/actions/security-guides/automatic-token-authentication it sounds like as long as we don't explicitly pass the `GITHUB_TOKEN` secret into the job, it'll not be able to use these permissions.

acamill:
Updating the `provider` to be reset for each test seems to fix the problems. LGTM. Need a squash

acamill:
Still too much instability to be able to add this as a auto run as it would break quite often.

It's mostly due to devnet :
- sometimes the oracle price is too far from the perp price, causing slippage errors (maybe the ci doesn't run the MM bot properly in these cases)
- sometimes the getBalance does not return the proper balance even after a succesful TX (still not sure why this happen, same connection. Seems to have some kind of latency, don't have this issue on local machine)

mschneider:
is there a part of the tests that doesnt need a working mango group and could just be run against the binary?
we do use the stub oracle for our own tests rather then pyth etc on devnet

acamill:
@microwavedcola1 This one is ready to go. I've remove the E2E test as it's still too janky, will make another PR.

ckamm (approved):
The tests look ok.

They currently fail because `BookSide::iter()` changed.

microwavedcola1 (approved):

microwavedcola1:
in principle I approve, but you might need to adapt code to work with latest master/main

acamill:
Yes updating tomorrow

acamill:
Updated with the planned patch branch v2.3.0 - I'm still wondering how we will proceed for deployment, as it would again need to be done in sync. Let me know your thoughs, cheers.

acamill:
(seeing https://github.com/blockworks-foundation/mango-v3/pull/144 , I believe this CI will keep UXD released version as checkout)

mschneider:
no need to deploy in sync, the current data layout simply has time in force always set to 0

acamill:
> no need to deploy in sync, the current data layout simply has time in force always set to 0

This will break without us updating I believe? Post audit we will use place_order_v2 to remove this composability weakness 

Screenshot 2022-02-22 at 19 56 16

acamill:
> no need to deploy in sync, the current data layout simply has time in force always set to 0
Screenshot 2022-02-23 at 09 07 20

This will break if we don't update on our side. And if we update before you it will break too.

This will be remove after audit for place_perp_order_v2, keeping minimal change for now

acamill:
Looks good to merge

PR #115, opened 2022-01-14
HealthCache: improvements for external use

ckamm:
This makes it easier to use the HealthCache externally, which I use for the liquidatable-accounts-feed.

### DIFF
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 no-entrypoint = []
 test-bpf = []
 devnet = []
+client = ["no-entrypoint"]
 
 [dependencies]
 solana-program = "^1.8.1"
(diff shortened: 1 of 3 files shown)


ckamm:
This changes the semantics slightly: all open orders accounts that are passed will be attempted to be loaded, even if their market is not in_margin_basket.

dafyddd:
Ok I'm going to push out v3.3.0 to mainnet, and I think this requires a bit more review. So maybe you can give this PR to main as per the new branch rules and I can merge it in their without deploying to mainnet

dafyddd:
Do you think there's something lost in readability from the T: Deref<..> instead of &OpenOrders or Ref<OpenOrders>? Do you have an example of where you're using this function from elsewhere?

ckamm:
Ok, will target `main` instead.

Yes, there's more cognitive overhead with being generic over `T: Deref<...>`. I want it to work with plain `&OpenOrders` because that's what I end up with in https://github.com/blockworks-foundation/liquidatable-accounts-feed/blob/master/src/main.rs#L148.

In mango-v3 we likely always have `Ref<OpenOrders->` because that's what we get from `AccountInfo`.

If it seems too confusing, I'd also be open to making it separate functions.

dafyddd:
Nah, not that confusing. If there's a legitimate use case no problem

dafyddd:
Can you reopen PR but so it's pointing at main and conflicts are resolved?

PR #119, opened 2022-01-21
add payer accounts to pda-creation instructions

ckamm:
Add a new optional account to
- create_mango_account
- create_spot_open_orders
that pays for the creation of the new account.

If you want to compose on mango and have a PDA account as an owner, it's
impossible to also use it as a payer because accounts with data cannot
pay for PDA creation.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -836,12 +836,13 @@ pub enum MangoInstruction {
 
     /// Create a PDA mango account for a user
     ///
-    /// Accounts expected by this instruction (4):
+    /// Accounts expected by this instruction (5):
     ///
     /// 0. `[writable]` mango_group_ai - MangoGroup that this mango account is for
     /// 1. `[writable]` mango_account_ai - the mango account data
     /// 2. `[signer]` owner_ai - Solana account of owner of the mango account
     /// 3. `[]` system_prog_ai - System program
+    /// 4. `[signer, writable]` payer_ai - pays for the PDA creation
     CreateMangoAccount {
         account_num: u64,
     },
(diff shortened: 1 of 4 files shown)


program/tests/test_create_account.rs
```diff
@@ -0,0 +1,60 @@
...
+    //
+    // paid for by separate payer
+    //
+    let payer = Keypair::from_base58_string(&test.users[1].to_base58_string());
+    let payer_lamports = test.get_lamport_balance(payer.pubkey()).await;
+    let owner_lamports = test.get_lamport_balance(test.users[0].pubkey()).await;
+    let account1_pk = test.create_mango_account(mango_group_pk, 0, 1, Some(&payer)).await;
+    assert_eq!(test.get_lamport_balance(test.users[0].pubkey()).await, owner_lamports);
+    assert!(test.get_lamport_balance(payer.pubkey()).await < payer_lamports);
+    assert!(account0_pk != account1_pk);
```
microwavedcola1, line 48: nitpick: can we use another payer than (test.users[0].pubkey()) assuming thats also the owner, and then add an assert that mango accounts owner is the owner and not the payer?

ckamm: The owner is `users[0]` and the payer is `users[1]`. I'll add the owner assert to make that clearer.

microwavedcola1 (approved):

program/src/instruction.rs
```diff
@@ -1539,13 +1541,15 @@ pub fn create_mango_account(
     mango_account_pk: &Pubkey,
     owner_pk: &Pubkey,
     system_prog_pk: &Pubkey,
+    payer_pk: &Pubkey,
```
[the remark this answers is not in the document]
dafyddd: Nevermind, only breaking if they upgrade the mango library in which case it will fail to compile.

PR #121, opened 2022-01-26
fix cancel_all bug

dafyddd:


program/src/matching.rs
```diff
@@ -1407,15 +1407,19 @@ impl<'a> Book<'a> {
...
                 Some(&my_highest_bid) => {
                     if bid.key > my_highest_bid {
                         cuml_bids += bid.quantity;
+                        curr = iter.next();
                     } else if bid.key == my_highest_bid {
                         bids_and_sizes.push((bid.key, cuml_bids));
                         my_bids.pop();
+                        curr = iter.next();
                     } else {
                         cuml_bids += bid.quantity;
```
ckamm, line 1424: I think it will now count `bid.quantity` twice

ckamm (approved):

microwavedcola1 (approved):

PR #124, opened 2022-01-31
Acam/place perp order v2

acamill:
Not sure about the destination.

No test, just a first draft @dafyddd let me know if anything sketchy

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -237,6 +237,31 @@ pub enum MangoInstruction {
         reduce_only: bool,
     },
 
+    /// Place an order on a perp market Improved over PlacePerpOrder
+    /// by reducing the tx size, and using max base/quote.
+    /// 
+    /// For OrderTypes Limit, PostOnly, PostOnlySlide the 
+    /// `max_base_quantity` will simply be `quantity`, and the 
+    /// `limit_price` be `price`.
+    /// 
+    /// The max parameters are useful when using instantaneous
+    /// OrderTypes (IoC or Market).
+    /// Goal is to allow orders with thresholds, without knowing
+    /// the quantity nor price in advance. 
+    /// It will check wether it fits within the provided `limit_price`,
+    /// It will optionally checks wether it's fully filled for IoC.
+    PlacePerpOrderV2 {
+        limit_price: i64,
+        max_base_quantity: i64,
+        max_quote_quantity: i64,
+        client_order_id: u64,
+        side: Side,
+        /// Can be 0 -> LIMIT, 1 -> IOC, 2 -> PostOnly
+        order_type: OrderType,
+        /// Optional to be backward compatible; default false
+        reduce_only: bool,
+    },
+
     CancelPerpOrderByClientId {
         client_order_id: u64,
         invalid_id_ok: bool,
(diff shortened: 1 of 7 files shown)


program/src/matching.rs
```diff
@@ -271,6 +271,16 @@ pub struct BookSideIter<'a> {
     right: usize,
 }
 
+pub struct Order {
+    // The quantity, in base_lot
+    pub quantity: i64,
+    // Marginal Price, the price to place the order at, in quote (per base_lot)
+    pub price: i64,
+    // The resulting total amount that will be spent, in quote_lot (without fees)
+    pub size: i64,
```
ckamm, line 280: size currently unused?

program/src/processor.rs
```diff
@@ -2479,6 +2479,167 @@ impl Processor {
...
+                }))
+            }
+            OrderType::ImmediateOrCancel | OrderType::Market => {
+                // Goes through the order book and find the price of quantity for the order
+                let depth_limit = 10;
+                if max_base_quantity > 0 {
+                    book.get_best_order_for_base_quantity(side, max_base_quantity, depth_limit);
+                }
+                if max_quote_quantity > 0 {
+                    book.get_best_order_for_quote_amount(side, max_quote_quantity, depth_limit);
```
ckamm, line 2586: The return values here and above are unused?

What's the difference between `get_best_order_for_base_quantity()` and using an extreme `limit_price` with the previous order matching?

Is the idea that one can combine `max_base_quantity` and `max_quote_quantity` constraints?

program/src/instruction.rs
```diff
@@ -237,6 +237,31 @@ pub enum MangoInstruction {
...
+    /// `max_base_quantity` will simply be `quantity`, and the
+    /// `limit_price` be `price`.
+    ///
+    /// The max parameters are useful when using instantaneous
+    /// OrderTypes (IoC or Market).
+    /// Goal is to allow orders with thresholds, without knowing
+    /// the quantity nor price in advance.
+    /// It will check wether it fits within the provided `limit_price`,
+    /// It will optionally checks wether it's fully filled for IoC.
+    PlacePerpOrderV2 {
```
ckamm, line 253: note: I'll also make a PlacePerpOrderV2 for the time-in-force feature. Let's get both new features into that new instruction.

program/src/matching.rs
acamill: Yes size is not used anymore, can be removed

program/src/processor.rs
```diff
@@ -2479,6 +2479,167 @@ impl Processor {
...
+                    price: limit_price,
+                    size: 0,
+                    side,
+                }))
+            }
+            OrderType::ImmediateOrCancel | OrderType::Market => {
+                // Goes through the order book and find the price of quantity for the order
+                let depth_limit = 10;
+                if max_base_quantity > 0 {
+                    book.get_best_order_for_base_quantity(side, max_base_quantity, depth_limit);
```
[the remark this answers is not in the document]
acamill: I was trying to do a first port of existing logic and then refine along the way. It's still very rough, but yes can probably fit there better.

program/src/processor.rs
acamill: No these constraints are independants. 

On my side I was using these two function to calculate an order before it happen to see how much collateral need to be deposited on mango and estimate other things.

I then saw the interface from 01 add these parameters in they equivalent of "place_perp_order" so I gave it a go, but it's probably gibberish as it is.

acamill:
Code deprecated from our side, and @ckamm working on implementing the max_quote_lot version after discussions today

PR #125, opened 2022-02-01
CI: Soteria performance improvements

silas-x:
PR contains changes that reduces the job run time significantly based on:

1. Updating toolchain without running `build cargo-bpf`
2. Caching Solana binaries and build deps

Job is set to pass until completion of audit/finding review


silas-x:
Merging as reviewed/tested in voter-stake-registry to reduce test time for other jobs

PR #126, opened 2022-02-02
Add perp balances logging

nicholasgodfreyclarke:
Allows for more efficient/robust queries.

### DIFF
--- a/mango-logs/src/lib.rs
+++ b/mango-logs/src/lib.rs
@@ -238,11 +238,11 @@ pub struct PerpBalanceLog {
     pub mango_account: Pubkey,
     pub market_index: u64, // IDL doesn't support usize
     pub base_position: i64,
-    pub quote_position: i128, // I80F48
-    pub long_settled_funding: i128, // I80F48
+    pub quote_position: i128,        // I80F48
+    pub long_settled_funding: i128,  // I80F48
     pub short_settled_funding: i128, // I80F48
     pub mngo_accrued: u64,
 
-    pub long_funding: i128, // I80F48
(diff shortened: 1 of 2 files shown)


nicholasgodfreyclarke:
Motivation for this PR:
In order to calculate the value of an account's perp position (for account equity, pnl, etc) we need to know the base position, quote position and mngo accrued for each perp account. This was previously done by summing all the changes in base/quote to the present however a more efficient method would be to simply log the current balance each time they are changed (also would not be at risk of failing to exactly replicate the current balance using a more complicate query in the DB). We can also simplify the logic by logging the current long and short funding when a change occurs (these are also required to obtain the final quote position).

To this end we log the current balances whenever a change in base/quote/mngo accrued occurs in the following instructions:

- ConsumeEvents
- SettlePnl
- LiquidateTokenAndPerp
- LiquidatePerpMarket
- SettleFees
- ResolvePerpBankruptcy
- RedeemMngo

microwavedcola1:
nitpick: @nicholasgodfreyclarke  just to make any of the the *CancelPerpOrders* instructions are not relevant, or?

microwavedcola1 (approved):

nicholasgodfreyclarke:
@microwavedcola1 thanks - that's a good point. I assume you're referring to mango being accrued in cancel perp orders?

My original intention was just to log base/quote balances.
So given the sheer number of cancel perp orders instructions - I'm thinking I should simplify this logging and remove the *mngo_accrued* field so it will log base/quote changes only. Then I will also remove the logging in RedeemMngo.
It's relatively straightforward to build up the accured mango balances from the existing logs.

PR #127, opened 2022-02-04
Dd/pyth status

dafyddd:


program/src/processor.rs
```diff
@@ -6216,10 +6214,7 @@ fn read_oracle(
             let conf = I80F48::from_num(price_account.agg.conf).checked_div(value).unwrap();
 
             #[cfg(not(feature = "devnet"))]
-            if price_account.agg.status != PriceStatus::Trading {
-                msg!("Pyth status invalid: {}", price_account.agg.status as u8);
-                return Err(throw_err!(MangoErrorCode::InvalidOraclePrice));
-            } else if conf > PYTH_CONF_FILTER {
```
microwavedcola1, line 6222: could you add a comment as to why unknown status is fine?

microwavedcola1 (approved):

program/src/processor.rs
dafyddd: done

PR #129, opened 2022-02-09
Add reflinks

ckamm:


ckamm:
I'd squash-merge this branch because the history is confusing (seems to have some cherry-picked `main` commits, several commits have the exact same commit message but different diffs etc)

program/src/matching.rs
```diff
@@ -999,12 +1024,21 @@ impl<'a> Book<'a> {
...
+        let info = &mango_group.perp_markets[market_index];
+        if post_allowed {
+            let native_price = market.lot_to_native_price(price);
+            if native_price.checked_div(oracle_price).unwrap() > info.maint_liab_weight {
+                msg!("Posting on book disallowed due to price limits");
+                post_allowed = false;
+            }
         }
 
+        // referral fee related variables
```
ckamm, line 1036: maybe add a note: "computed lazily to save cu on average"

program/src/matching.rs
```diff
@@ -1049,7 +1101,7 @@ impl<'a> Book<'a> {
                 *mango_account_pk,
                 order_id,
                 client_order_id,
-                info.taker_fee,
+                info.taker_fee + ref_fee_rate.unwrap(),
```
ckamm, line 1104: self note: this is used solely for logging

program/src/matching.rs
```diff
@@ -1610,3 +1782,46 @@ impl<'a> Book<'a> {
...
+    referrer_mango_account_ai: Option<&'a AccountInfo>,
+    now_ts: u64,
+) -> MangoResult<(I80F48, Option<RefMut<'a, MangoAccount>>)> {
+    let mngo_index = match mango_group.find_token_index(&mngo_token::id()) {
+        None => return Ok((ZERO_I80F48, None)),
+        Some(i) => i,
+    };
+
+    let mngo_cache = &mango_cache.root_bank_cache[mngo_index];
+    let mngo_deposits = mango_account.get_native_deposit(mngo_cache, mngo_index)?;
```
ckamm, line 1801: is this guaranteed to produce 0 if the users doesn't have any MNGO and the cache is potentially out of date?

program/src/matching.rs
```diff
@@ -1610,3 +1782,46 @@ impl<'a> Book<'a> {
...
+    let ref_mngo_req = I80F48::from_num(mango_group.ref_mngo_required);
+    if mngo_deposits >= ref_mngo_req {
+        return Ok((ZERO_I80F48, None));
+    } else if let Some(referrer_mango_account_ai) = referrer_mango_account_ai {
+        if let Ok(referrer_mango_account) =
+            MangoAccount::load_mut_checked(referrer_mango_account_ai, program_id, mango_group_pk)
+        {
+            // Need to check if it's valid because user may not have mngo in active assets
+            mngo_cache.check_valid(mango_group, now_ts)?;
```
ckamm, line 1810: I was wondering if it's intentional that an error from loading the referrer's MangoAccount is ignored above, but an error in this line or the get_native_deposit() below would be critical. It makes sense though: we don't want users to be affected when referrer closes their account.

Maybe add comments about this?

program/src/matching.rs
```diff
@@ -1610,3 +1782,46 @@ impl<'a> Book<'a> {
...
+                && !referrer_mango_account.being_liquidated
+                && ref_mngo_deposits >= ref_mngo_req
+            {
+                return Ok((
+                    I80F48::from_num(mango_group.ref_share_centibps) / CENTIBPS_PER_UNIT,
+                    Some(referrer_mango_account),
+                ));
+            }
+        }
+        // If referrer_mango_account is invalid, just treat it as if it doesn't exist
```
ckamm, line 1824: minor: I would move this above the `if`

program/src/processor.rs
```diff
@@ -5673,6 +5668,150 @@ impl Processor {
...
+        check!(
+            &referrer_mango_account.owner == owner_ai.key
+                || &referrer_mango_account.delegate == owner_ai.key,
+            MangoErrorCode::InvalidOwner
+        )?;
+        check!(owner_ai.is_signer, MangoErrorCode::InvalidSignerKey)?;
+
+        // referrer_id_record must be empty; cannot be transferred
+        check!(referrer_id_record_ai.data_is_empty(), MangoErrorCode::InvalidAccount)?;
+        let referrer_record_seeds: &[&[u8]] = &[&mango_group_ai.key.as_ref(), &referrer_id];
```
ckamm, line 5796: important: Add a fixed identifier between mango_group_ai and referrer_id. For example the seed `[mango_group_ai, "DustAccount"]` is already taken.

program/src/state.rs
```diff
@@ -2493,3 +2499,56 @@ impl AdvancedOrders {
         Ok(state)
     }
 }
+
+/// Store the referrer's mango account
+#[derive(Copy, Clone, Pod, Loadable)]
+#[repr(C)]
+pub struct ReferrerMemory {
+    pub referrer_mango_account: Pubkey,
```
ckamm, line 2507: add one byte for versioning, just in case (or maybe Metadata)

program/src/state.rs
```diff
@@ -2493,3 +2499,56 @@ impl AdvancedOrders {
...
+        check_eq!(account.owner, program_id, MangoErrorCode::InvalidOwner)?;
+        Ok(Self::load(account)?)
+    }
+}
+
+/// Register the referrer's id to be used in the URL
+#[derive(Copy, Clone, Pod, Loadable)]
+#[repr(C)]
+pub struct ReferrerIdRecord {
+    pub referrer_mango_account: Pubkey,
```
ckamm, line 2533: one byte for versioning

ckamm:
Note that one of the tests hits a MathError now:
```
thread 'test_perp_trigger_orders_health' panicked at 'order triggered: TransactionError(InstructionError(0, Custom(6)))', program/tests/test_perp_trigger_orders.rs:336:10
```
I'll look at that.

program/src/instruction.rs
```diff
@@ -925,6 +927,47 @@ pub enum MangoInstruction {
...
+    /// 0. `[]` mango_group_ai - MangoGroup that this mango account is for
+    /// 1. `[]` mango_account_ai - MangoAccount of the referred
+    /// 2. `[signer]` owner_ai - MangoAccount owner or delegate
+    /// 3. `[writable]` referrer_memory_ai - ReferrerMemory struct; will be initialized if required
+    /// 4. `[]` referrer_mango_account_ai - referrer's MangoAccount
+    /// 5. `[signer, writable]` payer_ai - payer for PDA; can be same as owner
+    /// 6. `[]` system_prog_ai - System program
+    SetReferrerMemory,
+
+    /// ssociate the referrer's MangoAccount with a human readable `referrer_id` which can be used
```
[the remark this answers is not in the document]
dafyddd: done

program/src/matching.rs
dafyddd: Yes. If the user's MNGO deposit is non-zero then the rootbank cache will be checked already in `place_perp_order`. If  it's zero then cache may be out of date, but it doesn't matter because 0 * index = 0. 

I will add a comment here.

program/src/matching.rs
dafyddd: error in get_native_deposit would just be a checked mul error which should always be critical. Error in cache I suppose shouldn't be critical if referrer doesn't have any MNGO. But cache error shouldn't be happening frequently anyway

dafyddd: I could move the cache check into the final if block... small potatoes either way

ckamm:
The only remaining issue is breaking perp trigger orders. Probably need to add the "no referrer" case to `sim_new_bid/ask` to make simulated and actual execution match up.

program/src/state.rs
```diff
@@ -1102,7 +1103,32 @@ impl HealthCache {
...
+                None => info.taker_fee * taker_quote_native,
+                Some(mngo_index) => {
+                    let mngo_cache = &mango_cache.root_bank_cache[mngo_index];
+                    let mngo_deposits = mango_account.get_native_deposit(mngo_cache, mngo_index)?;
+                    let ref_mngo_req = I80F48::from_num(mango_group.ref_mngo_required);
+                    if mngo_deposits >= ref_mngo_req {
+                        info.taker_fee * taker_quote_native
+                    } else {
+                        // done in two steps to perfectly simulate what happens in `apply_taker_fees`
+                        info.taker_fee * taker_quote_native
```
ckamm, line 1121: minor: all three branches contain `info.taker_fee * taker_quote_native`, could be moved out

program/src/state.rs
```diff
@@ -1102,7 +1103,32 @@ impl HealthCache {
...
-        Ok(h + curr_perp_health - prev_perp_health)
+
+        // Apply taker fees; Assume no referrer
+        let taker_fees = if taker_quote != 0 {
+            let taker_quote_native =
+                I80F48::from_num(info.quote_lot_size.checked_mul(taker_quote.abs()).unwrap());
+            match mango_group.find_token_index(&mngo_token::id()) {
+                None => info.taker_fee * taker_quote_native,
+                Some(mngo_index) => {
+                    let mngo_cache = &mango_cache.root_bank_cache[mngo_index];
```
ckamm, line 1114: do we need to check cache liveness here?

dafyddd: Not necessary because if user had any MNGO it would be in his ActiveAssets and it would have been checked already

PR #130, opened 2022-02-10
Add "time in force" for perp orders

microwavedcola1:


ckamm:
So, orders with time_in_force have a maximum lifetime of 256 seconds. It's sad we don't have more bytes available, but sufficient for MMs.

Thinking about how this could go wrong:
- In a situation where one side of the book gets no orders for some time, while a lot of tif orders are posted to the other side those dead tif orders wouldn't be removed. That could push out valid non-tif orders! Potential attack is sending 1024 1-second orders at the top of the book to remove all competitors' orders?
- Tif orders that are far from the top of the book can be long-lived. If the top of the book has 23 live orders, followed by 1000 dead tif orders and then one live "backstop"-style order, a newly incoming order with a better price than the backstop would replace the backstop order and leave the dead tif orders in there.

ckamm:
Potential approaches:
- We still have bytes available in InnerNode. Maybe we could bookkeep the closest expiring timestamp of the left and right subtree? Then if a book side is full, we could cheaply remove some dead orders?

ckamm:
What I did to address the issues mentioned above is:
- Delete an expired order every time a new order is added to the same side of the book. That provides a basic level of expired-order removal and means that a new order will never push out a valid order while there are expired ones left to remove.
- When matching orders, delete up to 5 expired orders at the top of the book. This helps keeping the top of the book cleaned up, reducing average compute use.
- Tested compute needs when needing to skip over many expired orders: Skipping over an expired order is roughly 100 CU, deleting an expired order from the book is 1000 CU. That means that even if the book started with 500 expired orders, iterating over them would not make the instruction go over the limit compute.

program/src/matching.rs
```diff
@@ -54,7 +63,9 @@ pub struct InnerNode {
     pub key: i128,
     /// indexes into `BookSide::nodes`
     pub children: [NodeHandle; 2],
-    pub padding: [u8; NODE_SIZE - 32],
+    /// the lowest expiry timestamp for the left and right subtrees
```
microwavedcola1, line 66: nitpick: maybe add a comment to why this is maintained here

program/src/matching.rs
```diff
@@ -75,17 +87,30 @@ impl InnerNode {
         let crit_bit = (search_key & crit_bit_mask) != 0;
         (self.children[crit_bit as usize], crit_bit)
     }
+
+    /// The lowest timestamp at which one of the contained LeafNodes expires.
+    #[inline(always)]
+    pub fn lowest_expiry(&self) -> u64 {
```
microwavedcola1, line 93: nitpick: I was thinking if earliest vs latest are better than lowest vs highest, feel free to reject

program/src/matching.rs
```diff
@@ -658,6 +786,49 @@ impl BookSide {
...
+    ///
+    /// This function walks up the `stack` of parents and applies the change where the
+    /// previous child's `outdated_expiry` is replaced by `new_expiry`.
+    fn update_parent_lowest_expiry(
+        &mut self,
+        stack: &[(NodeHandle, bool)],
+        mut outdated_expiry: u64,
+        mut new_expiry: u64,
+    ) {
+        for (parent_h, crit_bit) in stack.iter().rev() {
```
microwavedcola1, line 800: nitpick: the stack been an array is misleading a bit, I was confused for a moment why we are iterating and then reversing ;)

program/src/matching.rs
```diff
@@ -1044,16 +1167,37 @@ impl<'a> Book<'a> {
...
+        for (best_ask_h, best_ask) in self.asks.iter_all_including_invalid() {
+            if rem_quantity <= 0 {
+                break;
+            }
+
+            if !best_ask.is_valid(now_ts) {
+                // Remove the order from the book unless we've done that enough
+                if number_of_dropped_expired_orders < DROP_EXPIRED_ORDER_LIMIT {
+                    number_of_dropped_expired_orders += 1;
+                    let event = OutEvent::new(
```
microwavedcola1, line 1187: question: should expired orders end up on queue? Update: I see that booted out orders are also on queue, so lgtm

microwavedcola1: still I wonder how does one figure out that is an expired event? should we create a new type? ExpiredEvent?

microwavedcola1: hmm OutEvent is also ok I guess

PR #132, opened 2022-02-11
take maker fees from perp market if FillEvent happened before v3.3.5 …

dafyddd:
…upgrade

### DIFF
--- a/program/src/queue.rs
+++ b/program/src/queue.rs
@@ -229,7 +229,8 @@ pub struct FillEvent {
     pub maker_slot: u8,
     pub maker_out: bool, // true if maker order quantity == 0
     pub version: u8,
-    pub padding: [u8; 3],
+    pub market_fees_applied: bool,
+    pub padding: [u8; 2],
     pub timestamp: u64,
     pub seq_num: usize, // note: usize same as u64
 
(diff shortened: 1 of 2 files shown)


program/src/queue.rs
```diff
@@ -229,7 +229,8 @@ pub struct FillEvent {
     pub maker_slot: u8,
     pub maker_out: bool, // true if maker order quantity == 0
     pub version: u8,
-    pub padding: [u8; 3],
+    pub market_fees_applied: bool,
+    pub padding: [u8; 2],
```
microwavedcola1, line 233: (side note: padding is a misleading name, reserve is better, this is not really padding to align, anyway we I will try to go with reserve naming in v4 for zero copy)

program/src/state.rs
```diff
@@ -1622,6 +1622,9 @@ impl MangoAccount {
         pa.change_base_position(perp_market, base_change);
         let quote = I80F48::from_num(perp_market.quote_lot_size.checked_mul(quote_change).unwrap());
         let fees = quote.abs() * fill.maker_fee;
+        if !fill.market_fees_applied {
```
microwavedcola1, line 1625: so iiuc, with this change you are accounting for those events which were about to be processed before deploying last night, or? havent all these events now already passed? are any left to enter this if branch?

microwavedcola1: update: of course, those which are still open orders, got it

microwavedcola1 (approved):

ckamm (approved):

PR #134, opened 2022-02-12
Fix memory issues due to event logging, increase consume event processing limit to 8

ckamm:
DON'T MERGE YET: I want to manually squash some of these commits

### DIFF
--- a/mango-logs/src/lib.rs
+++ b/mango-logs/src/lib.rs
@@ -1,13 +1,57 @@
 use anchor_lang::prelude::*;
+use anchor_lang::Discriminator;
+use std::io::Write;
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 
-// Log to Program Log with a prologue so transaction scraper knows following line is valid mango log
+/// Log to Program Log with a prologue so transaction scraper knows following line is valid mango log
+///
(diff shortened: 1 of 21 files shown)


mango-logs/src/lib.rs
```diff
@@ -1,15 +1,59 @@
 use anchor_lang::prelude::*;
+use anchor_lang::Discriminator;
+use std::io::Write;
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 
-// Log to Program Log with a prologue so transaction scraper knows following line is valid mango log
+/// Log to Program Log with a prologue so transaction scraper knows following line is valid mango log
+///
+/// Warning: This will allocate large buffers on the heap which will never be released. Since the max
```
microwavedcola1, line 8: nitpick: could you please add a small comment why will never be released for future readers unfamiliar with solana runtime

microwavedcola1:
@ckamm do you have cu usage before/after?

microwavedcola1 (approved):

ckamm:
@microwavedcola1 CU use:

consume one event
- before: 16024
- after:  13654 (-2370)

consume 8 events
- before: fails during 6th event, consuming 96240; 96240/6*8 = 128320
- after: 118571

dafyddd (approved):

PR #135, opened 2022-02-17
Interest Rate Calculation Helper Function

bonedaddy:
Pulls out the interest rate calculation logic from `RootBank::update_index` and makes it a helper function.

### DIFF
--- a/program/src/state.rs
+++ b/program/src/state.rs
@@ -26,7 +26,7 @@ use crate::error::{check_assert, MangoError, MangoErrorCode, MangoResult, Source
 use crate::ids::mngo_token;
 use crate::matching::{Book, LeafNode, OrderType, Side};
 use crate::queue::{EventQueue, EventType, FillEvent};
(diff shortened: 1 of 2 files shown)


ckamm:
@bonedaddy Looks good to me, thanks! Could you make the new helper functions `#[inline(always)]` to avoid any potential stack issues?

bonedaddy:
@ckamm added

ckamm (approved):

ckamm:
On CI:
- the cargo audit failures have already been dealt with on `main`
- Lint and test fails because of formatting, Run `cargo fmt`. I can't push to your branch, so you'll need to do it.
- the UXD composition test fails during setup, because it can't read the secret access token, not sure why that is - I'd not consider it a blocker

ckamm:
I've moved this to #157, squashed and applied cargo fmt. There we'll check if CI passes and then merge.

PR #139, opened 2022-02-18
Add a max_quote argument to PlacePerpOrder2

ckamm:
Ready to merge (with merge commit) once CI passes


program/src/matching.rs
```diff
@@ -1181,12 +1185,13 @@ impl<'a> Book<'a> {
...
         // and then applied after this loop.
-        let mut rem_quantity = quantity; // base lots (aka contracts)
+        let mut rem_base_quantity = max_base_quantity; // base lots (aka contracts)
+        let mut rem_quote_quantity = max_quote_quantity;
         let mut ask_changes: Vec<(NodeHandle, i64)> = vec![];
         let mut ask_deletes: Vec<i128> = vec![];
         let mut number_of_dropped_expired_orders = 0;
         for (best_ask_h, best_ask) in self.asks.iter_all_including_invalid() {
-            if rem_quantity <= 0 {
+            if rem_base_quantity <= 0 || rem_quote_quantity <= 0 {
```
ckamm, line 1194: This exit condition doesn't work, since `rem_quote_quantity` can stay >0 due to rounding effects.

program/src/matching.rs
```diff
@@ -302,10 +302,38 @@ impl AsRef<AnyNode> for LeafNode {
 #[repr(u8)]
 #[serde(into = "u8", try_from = "u8")]
 pub enum OrderType {
```
acamill, line 304: Helpful!

ckamm:
@acamill Does this look good to you? You can use `max_quote_quantity` to have a quote-based limit of how much orders will be taken.
- ImmediateOrCancel will not fail the ix if less than max_quote_quantity got taken. (similarly to how it does not fail the ix if less than max_base_quantity get taken)
- max_quote_quantity does not take fees into account.

acamill:
> @acamill Does this look good to you? You can use `max_quote_quantity` to have a quote-based limit of how much will be taken from orders on the book.
> 
> - ImmediateOrCancel will not fail the ix if less than max_quote_quantity got taken. (similarly to how it does not fail the ix if less than max_base_quantity gets taken)
> 
> - max_quote_quantity does not take fees into account.

That sounds great! I'm getting out of quarantine hotel now, but will review in the coming days (to learn from how you embedded it into existing logic)

Cheers

dafyddd (approved):
looks good. merge into main at your earliest convenience

PR #144, opened 2022-02-22
Add back deprecated BookSide::iter()

ckamm:
Feel free to merge after review

### DIFF
--- a/program/src/matching.rs
+++ b/program/src/matching.rs
@@ -436,6 +436,14 @@ impl<'a> Iterator for BookSideIter<'a> {
 }
 
 impl BookSide {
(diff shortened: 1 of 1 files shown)


acamill:
This is great, will prevent having to hotfix as far as I understand. 

Will still plan a proper change for the next update.

acamill:
@ckamm Would it be possible to make this unbreaking?
Screenshot 2022-02-23 at 08 46 00

acamill:
It's temporary and when the second audit finishes we will use place_perp_order_v2

dafyddd (approved):

PR #149, opened 2022-03-01
Implementing Option Native to Mango v3

godmodegalactus:
In this PR is implementation of initial mechanics mainly. 
1) Creating options markets.
2) Enabling users to write options which create new option tokens.
3) Enable token holders to exercise option i.e swap quote tokens for underlying tokens.
4) Enable option writer to get either quote tokens or underlying tokens after the expiry.
5) Implementing mechanics for european option where exercise time is limited to specific timeslot after expiry.
6) Implementing orderbook for options tokens.
7) Implementing methods so that user can trade option tokens on the orderbook.
8) Tests to test all the essential part of this mechanics.
9) Instruction which only exercise the option tokens if it result is in profit and add the difference of quote tokens into account.

This PR has implemented most of the mechanics which will enable the options market to work. I is missing following details which will need some feedback from the community, mango DAO and technical feasibility:

1) Integration of fees.
2) Taking options into account to calculate the health of a user.
3) Giving writers interest on the underlying tokens they have deposited.


mschneider:
Just had a quick read-through need to spend more time with this. Could you help us a little and give a bit of context: to risk management & interaction with mango's margin system:

1) how are options collateralized?
2) how are options positions liquidated?

godmodegalactus:
[Updated]

Implementation of new way to calculate margin for option users. Option holders margin will increase if their option tokens can be exercised for profit. Writer can recover their margin buy buying more option tokens.
For option token holders the margin will be calculated if price of underlying assets is greater than the strike price then we assume that user has exercised the options and directly add the tokens into their spot tokens and remove the quote token equivalent to strike price.

PR #150, opened 2022-03-02
feat: cpi feature for instructions

aiell0:
## Changelog
* Add CPI feature so instructions can be wrapped with Anchor's CPI features.


ckamm:
Could you explain what's going on here? Why is it helpful to have a "cpi" feature that's not used anywhere inside the mango program?

microwavedcola1:
afaics cpi is an anchor feature, https://github.com/project-serum/anchor/blob/master/lang/syn/src/codegen/program/cpi.rs#L108, and mango is not built using anchor

ckamm:
@aiell0 Would you be up for reducing this to just the switchboard upgrade?

aiell0:
@ckamm I removed the cpi portion and left switchboard in there.

ckamm (approved):

dafyddd:
There are so many changes to cargo lock. Are all of these necessary? And why does switchboard 0.1.36 work now but there's a comment about how switchboard 0.1.36 doesn't work

aiell0:
@dafyddd I will look into the Cargo.lock changes. As far as the switchboard comments, you'd have to go back for that I'm not the one that placed that in there. If there are really no big issues with it we should just ignore this.

PR #152, opened 2022-03-09
reward fills more than cancels or self trades

mschneider:


dafyddd:
Do you know under what circumstances it would be more profitable to take yourself (and pay the 1bp net fee)? I worry this will make all our volume stats a little bit more suspicious to people.

mschneider:
hourly volume in MNGO terms < 10**4 MNGO / hour

program/src/state.rs
```diff
@@ -1647,7 +1649,7 @@ impl MangoAccount {
                     0,
                     fill.maker_timestamp,
                     fill.timestamp,
-                    fill.quantity,
+                    fill.quantity.checked_shl(lm_qty_shift as u32).unwrap(),
```
ckamm, line 1652: Hmm, are you sure this will work? The quantity will be limited by the `max_depth_size`.

I think it may work better with the incentive function to reduce the quantity when rewarding cancels?

Also, how about using a mul instead of a shift?

dafyddd: This is a good point.

program/src/processor.rs
```diff
@@ -5937,6 +5944,28 @@ impl Processor {
...
+        let [
+            mango_group_ai, // write
+            admin_ai        // read, signer
+        ] = accounts;
+
+        let mut mango_group = MangoGroup::load_mut_checked(mango_group_ai, program_id)?;
+        check!(admin_ai.is_signer, MangoErrorCode::SignerNecessary)?;
+        check_eq!(admin_ai.key, &mango_group.admin, MangoErrorCode::InvalidAdminKey)?;
+
+        mango_group.lm_fill_qty_shift = lm_fill_qty_shift;
```
dafyddd, line 5965: should print the before and after for governance instructions so it's easier to see on gov ui what is changing and how

program/src/state.rs
dafyddd: One side effect is this will reward you more the longer you were on the book. I think fills and cancels should actually be treated separately. Or perhaps, a fill can be given plus 100 seconds?

PR #155, opened 2022-03-11
Dd/equity calc

dafyddd:
Useful for calculating equity without replicating all this code.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -6652,7 +6652,7 @@ fn invoke_transfer<'a>(
 }
 
 #[inline(never)]
-fn read_oracle(
+pub fn read_oracle(
     mango_group: &MangoGroup,
     token_index: usize,
     oracle_ai: &AccountInfo,
--- a/program/src/state.rs
+++ b/program/src/state.rs
@@ -939,7 +939,7 @@ impl HealthCache {
                         perp_market_info.init_asset_weight,
                         perp_market_info.init_liab_weight,
                     ),
-                    HealthType::Equity => (),
+                    HealthType::Equity => (ONE_I80F48, ONE_I80F48, ONE_I80F48, ONE_I80F48),
                 };
 
             if self.active_assets.spot[i] {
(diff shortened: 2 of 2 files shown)


program/src/state.rs
```diff
@@ -83,12 +83,13 @@ pub enum DataType {
...
 }
 
-const NUM_HEALTHS: usize = 2;
+const NUM_HEALTHS: usize = 3;
 #[repr(usize)]
 #[derive(Copy, Clone, IntoPrimitive, TryFromPrimitive)]
 pub enum HealthType {
     Maint,
     Init,
+    Equity,
```
microwavedcola1, line 92: Could we please add one line descriptions to all these variants?

microwavedcola1: Still not sure, what equity means in our context 😓

program/src/state.rs
```diff
@@ -778,7 +780,7 @@ pub struct HealthCache {
     quote: I80F48,
 
     /// This will be zero until update_health is called for the first time
-    health: [Option<I80F48>; 2],
```
ckamm, line 781: use the NUM_HEALTHS constant here

program/src/state.rs
ckamm: Looks like it's the USD value of assets - liabilities. (all weights are 1)

program/src/processor.rs
```diff
@@ -6652,7 +6652,7 @@ fn invoke_transfer<'a>(
 }
 
 #[inline(never)]
-fn read_oracle(
+pub fn read_oracle(
```
ckamm, line 6655: unrelated/accident?

ckamm (approved):

program/src/state.rs
dafyddd: good catch

program/src/processor.rs
dafyddd: We were using it in our onchain trading program which needed to read the oracle. It's unrelated to equity calc, but related to other programs importing mango lib

PR #156, opened 2022-03-11
double valid interval for PriceCache

dafyddd:
double valid interval for price cache--it's low risk and should help with the number of InvalidPriceCache we get during network instability

### DIFF
--- a/program/src/state.rs
+++ b/program/src/state.rs
@@ -599,6 +599,9 @@ pub struct PriceCache {
 
 impl PriceCache {
     pub fn check_valid(&self, mango_group: &MangoGroup, now_ts: u64) -> MangoResult<()> {
+        // Hack: explicitly double valid_interval as a quick fix to make Mango
+        // less likely to become unusable when solana reliability goes bad.
+        // There's currently no instruction to change the valid_interval.
         check!(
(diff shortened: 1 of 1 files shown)


mschneider:
probably need 20s given how unreliable the networks is, so 4x 5s ?

microwavedcola1:
What kind of manipulation becomes easier with this  change or max's suggstion?

program/src/state.rs
```diff
@@ -600,7 +600,7 @@ pub struct PriceCache {
 impl PriceCache {
     pub fn check_valid(&self, mango_group: &MangoGroup, now_ts: u64) -> MangoResult<()> {
         check!(
-            self.last_update >= now_ts - mango_group.valid_interval,
+            self.last_update >= now_ts - (2 * mango_group.valid_interval),
```
ckamm, line 603: Please add a comment, or some other hint that explains it, like
```
// Hack: explicitly double valid_interval as a quick fix to make Mango
// less likely to become unusable when solana reliability goes bad.
// There's currently no instruction to change the valid_interval.
```

ckamm (approved):

mschneider:
image

tx confirmation times during tps issues

PR #158, opened 2022-03-13
update anchor version and cargo deps and switchboard version

dafyddd:


dafyddd:
@acamill Is it necessary that we stay with solana program 1.9.1? can you update to 1.9.4?

acamill:
> @acamill Is it necessary that we stay with solana program 1.9.1? can you update to 1.9.4?

No we this is a mistake we don't require a specific version, was to follow the one you guys used

acamill:
I'll push a change for the CI to pass, no need to do any change to the onchain program if I understand correctly how dependencies work.

ckamm:
it looks like this is downgrading solana from 1.9.7 to 1.9.4, I wonder why that is

ckamm:
It's because `switchboard-utils` now pins the solana version:
```
[dependencies.solana-program]
version = "=1.9.4"
```
We'll need to ask them about that.

ckamm (approved):

PR #163, opened 2022-03-23
allow settle fees to silently fail

mschneider:
this is supposed to enable closing out a whole account in a single tx

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -3083,8 +3083,10 @@ impl Processor {
         let contract_size = mango_group.perp_markets[market_index].base_lot_size;
         let new_quote_pos = I80F48::from_num(-pa.base_position * contract_size) * price;
         let pnl: I80F48 = pa.quote_position - new_quote_pos;
-        check!(pnl.is_negative(), MangoErrorCode::Default)?;
-        check!(perp_market.fees_accrued.is_positive(), MangoErrorCode::Default)?;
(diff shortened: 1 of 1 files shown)


microwavedcola1 (approved):

program/src/processor.rs
```diff
@@ -3083,8 +3083,10 @@ impl Processor {
         let contract_size = mango_group.perp_markets[market_index].base_lot_size;
         let new_quote_pos = I80F48::from_num(-pa.base_position * contract_size) * price;
         let pnl: I80F48 = pa.quote_position - new_quote_pos;
-        check!(pnl.is_negative(), MangoErrorCode::Default)?;
-        check!(perp_market.fees_accrued.is_positive(), MangoErrorCode::Default)?;
+        // ignore these cases and fail silently so transactions can continue
+        if pnl.is_negative() || perp_market.fees_accrued.is_positive() {
```
dafyddd, line 3087: I think you want `if !(pnl.is_negative() && perp_market.fees_accrued.is_positive())`

mschneider: right

dafyddd: I think you should also print a `msg!` explaining why it's failing silently. We've used that standard elsewhere

microwavedcola1: hmm, good catch, embarrassing oversight from me, no change is small ;)

PR #168, opened 2022-04-14
add place_spot_order2 builder

microwavedcola1:
Signed-off-by: microwavedcola1 <microwavedcola@gmail.com>

There was no helper for composers, added that, acammm has already tested it, works fine.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -2563,7 +2563,6 @@ pub fn place_spot_order2(
     dex_signer_pk: &Pubkey,
     msrm_or_srm_vault_pk: &Pubkey,
     open_orders_pks: &[Pubkey],
-
     affected_market_open_orders_index: usize, // used to determine which of the open orders accounts should be passed in write
     order: serum_dex::instruction::NewOrderInstructionV3,
 ) -> Result<Instruction, ProgramError> {
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -2539,6 +2539,73 @@ pub fn place_spot_order(
     Ok(Instruction { program_id: *program_id, accounts, data })
 }
 
+pub fn place_spot_order2(
+    program_id: &Pubkey,
+    mango_group_pk: &Pubkey,
+    mango_account_pk: &Pubkey,
+    owner_pk: &Pubkey,
+    mango_cache_pk: &Pubkey,
+    dex_prog_pk: &Pubkey,
+    spot_market_pk: &Pubkey,
+    bids_pk: &Pubkey,
+    asks_pk: &Pubkey,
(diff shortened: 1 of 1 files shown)


program/src/instruction.rs
```diff
@@ -2539,6 +2539,73 @@ pub fn place_spot_order(
...
+        AccountMeta::new(*quote_node_bank_pk, false),
+        AccountMeta::new(*quote_vault_pk, false),
+        AccountMeta::new_readonly(spl_token::ID, false),
+        AccountMeta::new_readonly(*signer_pk, false),
+        AccountMeta::new_readonly(*dex_signer_pk, false),
+        AccountMeta::new_readonly(*msrm_or_srm_vault_pk, false),
+    ];
+
+    accounts.extend(open_orders_pks.iter().enumerate().map(|(i, pk)| {
+        if i == market_index {
```
ckamm, line 2596: This does not look right: the instruction expects an open orders account only for the markets that are in the margin basket. So if the user only is active on market 5, then they should pass a single oo account, the one for market 5.

`market_index` should probably be renamed to `affected_market_open_orders_index` or something like that?

ckamm (approved):

program/src/instruction.rs
```diff
@@ -2539,6 +2539,72 @@ pub fn place_spot_order(
...
+
+    accounts.extend(open_orders_pks.iter().enumerate().map(|(i, pk)| {
+        if i == affected_market_open_orders_index {
+            AccountMeta::new(*pk, false)
+        } else {
+            AccountMeta::new_readonly(*pk, false)
+        }
+    }));
+
+    let instr = MangoInstruction::PlaceSpotOrder { order };
```
dafyddd, line 2602: I believe this should be 
```rust
    let instr = MangoInstruction::PlaceSpotOrder2 { order };
```

program/src/instruction.rs
```diff
@@ -2539,6 +2539,72 @@ pub fn place_spot_order(
...
+    base_root_bank_pk: &Pubkey,
+    base_node_bank_pk: &Pubkey,
+    base_vault_pk: &Pubkey,
+    quote_root_bank_pk: &Pubkey,
+    quote_node_bank_pk: &Pubkey,
+    quote_vault_pk: &Pubkey,
+    signer_pk: &Pubkey,
+    dex_signer_pk: &Pubkey,
+    msrm_or_srm_vault_pk: &Pubkey,
+    open_orders_pks: &[Pubkey],
```
dafyddd, line 2565: nitpick: can add a comment saying you only need to pass in open_orders_pks that are in margin basket

dafyddd (approved):

PR #169, opened 2022-04-15
Add Associated Token Accounts owner check during withdraw

abdulwahed786:
Associated Token Accounts(ATA) ownership is transferable which is generally not normal

In case of a scenario where a user might have signed some malicious transactions in which the spammer had taken over the ownership of the user's ATA say-USDC, now if the user unknowingly tries to withdraw from the mango account the funds will get transferred to the USDC ATA owned by the spammer and not the user.

Below is the transaction in which I tried replicating the above scenario:  where its deterministic USDC ATA  is not owned by the user and tries withdrawing 5 USDC from mango

user wallet :  DC9wKR6F1PsLzjs1PikthtNaKtVfUeBCc4EUeHhsBLA1
spammer WALLET : 5yFs5rrUGqdwAww69DV6derLUZccs5zUQU6oe2BzUcop
user deterministic USDC ATA owner by spammer : 8nTLNSgkoFXpkLT6z4Y1hS6f49n3yaXruTTrCKmPCYK5

https://explorer.solana.com/tx/4dbGe7wxMPiQeVsEvceJCSZfDxD84M3pKzbykXY9nuw7ePEL5ApQ4tuzNcyTust7nNqfWJEPRX1FcewCUQY4qtVt?cluster=mainnet

### DIFF
--- a/program/src/error.rs
+++ b/program/src/error.rs
@@ -123,6 +123,8 @@ pub enum MangoErrorCode {
     InvalidOraclePrice,
     #[error("MangoErrorCode::MaxAccountsReached The maximum number of accounts for this group has been reached")]
     MaxAccountsReached,
+    #[error("MangoErrorCode::InvalidAccountOwner")]
+    InvalidAccountOwner,
 
     #[error("MangoErrorCode::Default Check the source code for more info")] // 40
     Default = u32::MAX_VALUE,
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -1319,6 +1319,8 @@ impl Processor {
             -withdraw,
         )?;
 
+        let token_account = Account::unpack(&token_account_ai.try_borrow_data()?)?;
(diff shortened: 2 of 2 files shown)


dafyddd:
I thought this would be good, but we also have intentions of one day allowing you to transfer from your MangoAccount. Anyway, we can do that with a withdraw then an spl transfer or maybe even a new ix. What do you think @tjshipe ?

abdulwahed786:
I don't think the MangoAccount ownership can be transferred like for ATA using setAuthority() instruction available on the AToken Program. couldn't find any such instruction on the mango program. 
image

abdulwahed786:
Although if this is a breaking change for some composers, then 
-  Maybe we should at least have a check on the MangoClient to check the ownership of the ATA before passing it to the Withdraw (such that the example I mentioned above doesn't happen with others) 
- Or have a new Instruction for composers to have them withdraw to a different ATA

PR #170, opened 2022-04-16
Dd/serum max ts

dafyddd:
Upgrade our serum-dex dependency and allow for max_ts feature in PlaceSpotOrder and PlaceSpotOrder2

Made a lot of changes to tests to make it work with new serum dex and also to speed it up

Everything works. Ready for merge


program/src/state.rs
```diff
@@ -2299,7 +2299,7 @@ pub fn load_market_state<'a>(
             from_bytes_mut(mid)
         });
 
-    state.check_flags()?;
+    state.check_flags(false)?;
```
tjshipe, line 2302: What does the false do here?

dafyddd: Serum added ability for Market to be disabled. In old version, default behavior was to error if Market was disabled. So by setting `allow_disabled` to false I preserve the old behavior.

tjshipe (approved):

PR #173, opened 2022-04-18
Load serum_dex dependency with a commit ref

jkbpvsc:
## Problem
Recently the mango fork of `serum_dex` was [merged](https://github.com/blockworks-foundation/serum-dex/commit/7f55a5ef5f7937b74381a3124021a261cd7d7283) with the upstream version 0.5.5. 
The main mango program is importing the `serum_dex` package with version `0.4.0` from a git repository.

If you delete Cargo.lock, the build will fail with 
```
cargo build-bpf
Failed to obtain package metadata: `cargo metadata` exited with an error:     Updating git repository `https://github.com/blockworks-foundation/serum-dex.git`
error: failed to select a version for the requirement `serum_dex = "^0.4.0"`
candidate versions found which didn't match: 0.5.5
location searched: Git repository https://github.com/blockworks-foundation/serum-dex.git
required by package `mango v3.4.3 (/Users/jakob/Projects/crypto/blockworks-foundation/mango-v3/program)`
```
because while Cargo.toml specifies version `0.4.0`, when importing from a git repo, the latest `0.5.5` from upstream is used.

### Fix
Build with `serum_dex 0.5.5` is breaking
```
error[E0063]: missing field `max_ts` in initializer of `NewOrderInstructionV3`
    --> program/src/instruction.rs:1621:10
     |
1621 |     Some(serum_dex::instruction::NewOrderInstructionV3 {
     |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `max_ts`

error[E0061]: this function takes 1 argument but 0 arguments were supplied
    --> program/src/state.rs:2302:11
     |
2302 |     state.check_flags()?;
     |           ^^^^^^^^^^^- supplied 0 arguments
     |           |
     |           expected 1 argument
     |
note: associated function defined here
    --> /Users/jakob/.cargo/git/checkouts/serum-dex-ac3ff2707f895b6e/7f55a5e/dex/src/state.rs:440:12
     |
440  |     pub fn check_flags(&self, allow_disabled: bool) -> DexResult {
```
 so as a quick fix, I removed the version from the `serum_dex` and fixated it to the [commit](https://github.com/blockworks-foundation/serum-dex/commit/3104f424ee38a415418a1cdef67970771f832857) before the upstream merge.

PR #175, opened 2022-04-19
Disable UXD CI until we open source

acamill:
Discussed on #173 
Screenshot 2022-04-19 at 13 13 18


acamill:
> Sounds good. I like the composability test. Would really like to not break UXD. Can't wait for the open source.

Yes I also like the idea too but right now it seems quite unsustainable with close source repo, don't want to abuse you guys hospitality ❤️

Let's just keep tight com until we release next version, as we are still running with some stranded mango logic

PR #178, opened 2022-04-19
Revert "Allow solana dependecies above `1.10` (#174)"

dafyddd:
This reverts commit a12a55947632764a31274c24a41af3aed68c944b.


dafyddd:
Sorry I accidentally merged, then reverted it. I'm going to be a bit careful about upgrading all the packages. Needs to be tested on devnet etc.

jkbpvsc:
Makes sense!

PR #186, opened 2022-05-12
V3.4.6

dafyddd:


microwavedcola1 (approved):

changelog.md
```diff
@@ -1,5 +1,11 @@
 # Mango Program Change Log
 
+## v3.4.6
+Deployed: | Slot:
+1. Force reduce only for perp market advanced order
+2. Allow LUNA-PERP bids to be as high as 9c
```
ckamm, line 6: spot too

program/src/matching.rs
```diff
@@ -1026,7 +1029,15 @@ impl<'a> Book<'a> {
         if post_allowed {
             // price limit check computed lazily to save CU on average
             let native_price = market.lot_to_native_price(price);
-            if native_price.checked_div(oracle_price).unwrap() > info.maint_liab_weight {
+
+            // Temporary hard coding LUNA price limit for bid to be below 10c.
+            // This is safe because it's already in reduce only mode
+            if market_index == LUNA_MARKET_INDEX && native_price >= market.lot_to_native_price(10) {
```
ckamm: Could you explain why this is needed? If the oracle is below 9 the normal price check would cover it. So this is to guard against weird oracle data?

program/src/processor.rs
```diff
@@ -2560,6 +2560,13 @@ impl Processor {
 
             if (side == Side::Bid && base_pos > 0) || (side == Side::Ask && base_pos < 0) {
                 0
+            } else if is_luna_market {
+                // Take into account outstanding open orders as well
+                let on_orders = match side {
+                    Side::Bid => mango_account.perp_accounts[market_index].bids_quantity,
```
ckamm, line 2566: Ok, I see base_pos already took perpmarket.taker_base and filled maker orders into account.

ckamm (approved):

program/src/matching.rs
ckamm: Ah, I think I get it. The existing price limit was too strict given the erratic oracle. This actually makes it more permissive.

program/src/processor.rs
```diff
@@ -2012,21 +2057,64 @@ impl Processor {
...
+            let order_quantity = order.max_coin_qty.get().checked_mul(coin_lot_size).unwrap();
+            match order_side {
+                serum_dex::matching::Side::Bid => {
+                    check!(
+                        net_pos.is_negative() && order_quantity <= net_pos.abs().to_num::<u64>(),
+                        MangoErrorCode::InvalidOrderInClosingMarket
+                    )?;
+
+                    (
+                        order.limit_price.get() < 10,
```
ckamm, line 2101: maybe we could compare `native_price` instead, then we don't need to double check coin_lots and pc_lots to see what the constant means.

dafyddd: isn't native price also confusing?

PR #189, opened 2022-05-23
allow passing in as many open orders as needed

mschneider:
Withdraw instruction size was too large for governance.
This change removes the requirement to pass default pks for spot open orders on withdraw, when the user doesn't have any.

### DIFF
--- a/program/src/state.rs
+++ b/program/src/state.rs
@@ -802,7 +802,7 @@ impl HealthCache {
         mango_group: &MangoGroup,
         mango_cache: &MangoCache,
         mango_account: &MangoAccount,
-        open_orders_ais: &[AccountInfo; MAX_PAIRS],
+        open_orders_ais: &[AccountInfo],
     ) -> MangoResult<()> {
         self.quote = mango_account.get_net(&mango_cache.root_bank_cache[QUOTE_INDEX], QUOTE_INDEX);
         for i in 0..mango_group.num_oracles {
@@ -811,7 +811,7 @@ impl HealthCache {
                     &mango_cache.root_bank_cache[i],
                     mango_cache.price_cache[i].price,
                     i,
-                    &if *open_orders_ais[i].key == Pubkey::default() {
+                    &if mango_account.spot_open_orders[i] == Pubkey::default() {
                         None
                     } else {
                         Some(load_open_orders(&open_orders_ais[i])?)
@@ -1524,7 +1524,7 @@ impl MangoAccount {
     pub fn check_open_orders(
         &self,
         mango_group: &MangoGroup,
-        open_orders_ais: &[AccountInfo; MAX_PAIRS],
+        open_orders_ais: &[AccountInfo],
     ) -> MangoResult {
         for i in 0..mango_group.num_oracles {
             if self.in_margin_basket[i] {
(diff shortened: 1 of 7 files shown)


program/src/state.rs
```diff
@@ -1524,7 +1524,7 @@ impl MangoAccount {
     pub fn check_open_orders(
         &self,
         mango_group: &MangoGroup,
-        open_orders_ais: &[AccountInfo; MAX_PAIRS],
+        open_orders_ais: &[AccountInfo],
     ) -> MangoResult {
         for i in 0..mango_group.num_oracles {
             if self.in_margin_basket[i] {
```
ckamm, line 1530: It looks like this would produce a bounds error when accessing `open_orders_ais[i]` for a token index that's in_margin_basket if not all open orders accounts are passed.

Maybe:
- use a separate index `j` for looking into `open_orders_ai`
- skip all `open_orders_ai[j] == Pubkey::default()` (to preserve the old behavior when an account is passed for each pair)
- if `spot_open_orders[i] == Pubkey::default()`: don't advance `j`

To make it possible to support the old case as well as the new case where you send only the open orders accounts that are needed?

program/src/state.rs
```diff
@@ -811,7 +811,7 @@ impl HealthCache {
                     &mango_cache.root_bank_cache[i],
                     mango_cache.price_cache[i].price,
                     i,
-                    &if *open_orders_ais[i].key == Pubkey::default() {
+                    &if mango_account.spot_open_orders[i] == Pubkey::default() {
                         None
                     } else {
                         Some(load_open_orders(&open_orders_ais[i])?)
```
ckamm, line 817: This would work if an account needs no open orders at all. But imagine they need the oo account for i=4 and pass only that. It looks like this would try to look up `open_orders_ais[4]`.

ckamm (changes requested):

program/src/state.rs
mschneider: yes, it only supports truncating the end right now, not leaving out spots in the middle

program/src/state.rs
mschneider: As discussed this creates a lot of issues with backwards compatibility as check_open_orders  and get_val method is used across a lot of different instructions and there are edge cases like open orders accounts being stored on the mango account even when a token is not in the margin basket.

mschneider:
Updated with a custom error code, although this one hides a bit of information (array index of access failure).

program/src/state.rs
```diff
@@ -802,20 +802,25 @@ impl HealthCache {
         mango_group: &MangoGroup,
         mango_cache: &MangoCache,
         mango_account: &MangoAccount,
-        open_orders_ais: &[AccountInfo; MAX_PAIRS],
+        open_orders_ais: &[AccountInfo],
     ) -> MangoResult<()> {
         self.quote = mango_account.get_net(&mango_cache.root_bank_cache[QUOTE_INDEX], QUOTE_INDEX);
         for i in 0..mango_group.num_oracles {
             if self.active_assets.spot[i] {
+                let oo: MangoResult<Option<Ref<serum_dex::state::OpenOrders>>> = if mango_account
```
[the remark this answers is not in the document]
mschneider: need a Result wrapper to bubble up check failure, previous code only had an option return type, but we need 3 cases: Some, None, Error

mschneider:
based on feedback from daffy deprecated in favor of a solution using another instruction, as we anyways want to add more withdraw features: https://github.com/blockworks-foundation/mango-v3/pull/195

PR #190, opened 2022-05-24
Delist Market

dafyddd:
This is still a work in progress and all functions have not been implemented and safety has not been verified.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -3049,10 +3049,11 @@ pub fn liquidate_delisting_token(
     liab_vault_pk: &Pubkey,
     liqee_liab_token_account_pk: &Pubkey,
     liqor_liab_token_account_pk: &Pubkey,
+    liqee_open_orders_pks: &[Pubkey],
+    liqor_open_orders_pks: &[Pubkey],
     signer_pk: &Pubkey,
-    token_program_pk: &Pubkey,
 ) -> Result<Instruction, ProgramError> {
-    let accounts = vec![
+    let mut accounts = vec![
         AccountMeta::new_readonly(*mango_group_pk, false),
         AccountMeta::new_readonly(*mango_cache_pk, false),
         AccountMeta::new(*liqee_mango_account_pk, false),
@@ -3066,8 +3067,11 @@ pub fn liquidate_delisting_token(
         AccountMeta::new(*liqee_liab_token_account_pk, false),
         AccountMeta::new(*liqor_liab_token_account_pk, false),
         AccountMeta::new_readonly(*signer_pk, false),
-        AccountMeta::new_readonly(*token_program_pk, false),
+        AccountMeta::new_readonly(spl_token::ID, false),
     ];
+
+    accounts.extend(liqee_open_orders_pks.iter().map(|pk| AccountMeta::new_readonly(*pk, false)));
+    accounts.extend(liqor_open_orders_pks.iter().map(|pk| AccountMeta::new_readonly(*pk, false)));
 
     let instr = MangoInstruction::LiquidateDelistingToken;
     let data = instr.pack();
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -1121,24 +1121,27 @@ pub enum MangoInstruction {
     RemoveOracle,
 
     /// Resolve deposits/borrows of a delisting token for an account
-    /// Accounts expected by this instruction (14 + 2 * MAX_PAIRS)
+    /// Accounts expected by this instruction (15 + 2 * MAX_PAIRS)
     /// 0. `[]` mango_group_ai
     /// 1. `[]` mango_cache_ai
-    /// 2. `[writable]` liqee_mango_account_ai
-    /// 3. `[writable]` liqor_mango_account_ai
-    /// 4. `[signer]` liqor_ai
-    /// 5. `[]` asset_root_bank_ai
-    /// 6. `[writable]` asset_node_bank_ai
-    /// 7. `[]` liab_root_bank_ai
-    /// 8. `[writable]` liab_node_bank_ai
-    /// 9. `[writable]` liab_vault_ai
-    /// 10. `[writable]` liqee_liab_token_account_ai
-    /// 11. `[writable]` liqor_liab_token_account_ai
-    /// 12. `[]` signer_ai
-    /// 13. `[]` token_prog_ai
-    /// 13... `[]` liqee_open_orders_ais - Liqee open orders accs
-    /// 13+MAX_PAIRS... `[]` liqor_open_orders_ais - Liqor open orders accs
-    LiquidateDelistingToken,
+    /// 2. `[writable]` dust_account_ai
+    /// 3. `[writable]` liqee_mango_account_ai
+    /// 4. `[writable]` liqor_mango_account_ai
+    /// 5. `[signer]` liqor_ai
+    /// 6. `[]` asset_root_bank_ai
+    /// 7. `[writable]` asset_node_bank_ai
+    /// 8. `[]` liab_root_bank_ai
+    /// 9. `[writable]` liab_node_bank_ai
+    /// 10. `[writable]` liab_vault_ai
+    /// 11. `[writable]` liqee_liab_token_account_ai
+    /// 12. `[writable]` liqor_liab_token_account_ai
+    /// 13. `[]` signer_ai
+    /// 14. `[]` token_prog_ai
+    /// 14... `[]` liqee_open_orders_ais - Liqee open orders accs
+    /// 14+MAX_PAIRS... `[]` liqor_open_orders_ais - Liqor open orders accs
+    LiquidateDelistingToken {
+        max_liquidate_amount: u64,
+    },
 
     /// Force settle a user's perp positions
     /// Accounts expected by this instruction (10)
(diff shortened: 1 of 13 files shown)


program/src/state.rs
```diff
@@ -218,7 +254,10 @@ pub struct MangoGroup {
     pub ref_surcharge_centibps: u32, // 100
     pub ref_share_centibps: u32,     // 80 (must be less than surcharge)
     pub ref_mngo_required: u64,
-    pub padding: [u8; 8], // padding used for future expansions
+
+    pub oracle_inactive_bits: u16,
```
ckamm, line 258: probably unused, can be dropped

ckamm:
create_perp_market should probably set the perp_market_mode to active

program/src/processor.rs
```diff
@@ -1399,6 +1299,14 @@ impl Processor {
             .find_spot_market_index(spot_market_ai.key)
             .ok_or(throw_err!(MangoErrorCode::InvalidMarket))?;
 
+        // Don't allow initialization of spot open orders if market is in ForceClose or Swapping
+        // Don't need to check MarketMode::Inactive because spot market is zero key in that case
+        let mode = mango_group.tokens[market_index].spot_market_mode;
+        check!(
+            !(mode == MarketMode::ForceCloseOnly || mode == MarketMode::SwappingSpotMarket),
```
ckamm, line 1306: possibly there could be `spot_market_mode.is_reduce_only()` and `spot_market_mode.allow_new_open_orders()`

program/src/processor.rs
```diff
@@ -3008,6 +2947,15 @@ impl Processor {
 
         let market_index = mango_group.find_perp_market_index(perp_market_ai.key).unwrap();
 
+        // Owner signature not necessary if market is in ForceClose
+        let mode = mango_group.tokens[market_index].perp_market_mode;
+        if mode == MarketMode::ForceCloseOnly {
+            // To guard against possibility delegate is zero key and that's passed in
+            check!(owner_ai.key != &Pubkey::default(), MangoErrorCode::InvalidOwner)?;
```
ckamm, line 2954: I don't understand this check, can you explain? `owner_ai` is either the owner or the delegate. Why is passing Pubkey::default() bad?

program/src/processor.rs
```diff
@@ -2770,7 +2710,7 @@ impl Processor {
             now_ts,
             referrer_mango_account_ai,
             limit,
-            is_luna_market,
+            is_market_closing,
```
riordanp, line 2713: This will limit new bids for all closing markets to 10c, not just LUNA. Can't imagine this is desired, so will re-add is_luna_market and pass that instead

program/src/processor.rs
```diff
@@ -2599,7 +2536,7 @@ impl Processor {
             now_ts,
             referrer_mango_account_ai,
             u8::MAX,
-            is_luna_market,
+            is_market_closing,
```
riordanp, line 2539: This will limit new bids for all closing markets to 10c, not just LUNA. Can't imagine this is desired, so will re-add is_luna_market and pass that instead

ckamm:
Just for clarity and extra safety, I think we should always set all data in accounts to zero when we set `metadata.initialized = false`

program/src/processor.rs
dafyddd: You're right this check is not needed

program/src/processor.rs
```diff
@@ -6365,57 +6304,901 @@ impl Processor {
...
+        };
+
+        mango_group.tokens[market_index].mint = Pubkey::default();
+        mango_group.tokens[market_index].root_bank = Pubkey::default();
+        mango_group.tokens[market_index].spot_market_mode = MarketMode::Inactive;
+        Ok(())
+    }
+
+    #[inline(never)]
+    /// For a market in `SwappingSpotMarket` where we know all the OpenOrders have been closed,
```
ckamm, line 6502: But we don't actually know that all openorders are closed, right? The market mode change doesn't check that, and we don't check it here either. And in fact there's no way to check for that, right?

ckamm (commented):
First part. Still need to read remove_perp, remove_oracle, liquidate, force_settle.

program/src/state.rs
```diff
@@ -286,6 +349,12 @@ impl MangoGroup {
             }
         }
     }
+
+    pub fn is_spot_market_active(&self, market_index: usize) -> bool {
+        !self.spot_markets[market_index].is_empty()
+            && (self.tokens[market_index].spot_market_mode == MarketMode::Active
+                || self.tokens[market_index].spot_market_mode == MarketMode::Default)
```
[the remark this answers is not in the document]
riordanp: don't think this is even used, can probably just delete it

program/src/instruction.rs
```diff
@@ -1066,6 +1066,91 @@ pub enum MangoInstruction {
...
+    /// 7. `[writable]` asset_node_bank_ai
+    /// 8. `[]` liab_root_bank_ai
+    /// 9. `[writable]` liab_node_bank_ai
+    /// 10. `[writable]` liab_vault_ai
+    /// 11. `[writable]` liqee_liab_token_account_ai
+    /// 12. `[writable]` liqor_liab_token_account_ai
+    /// 13. `[]` signer_ai
+    /// 14. `[]` token_prog_ai
+    /// 14... `[]` liqee_open_orders_ais - Liqee open orders accs
+    /// 14+MAX_PAIRS... `[]` liqor_open_orders_ais - Liqor open orders accs
```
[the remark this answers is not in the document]
riordanp: Will try the same method as Max's Withdraw2 so the liqor can pass only the keys that are inMarginBasket, but liqors with lots of OOs will fail to submit the transaction

program/src/processor.rs
```diff
@@ -6365,58 +6304,902 @@ impl Processor {
...
+                delist_index,
+                delist_transfer,
+            )?;
+            delist_net = liqee_ma.get_net(delist_root_bank_cache, delist_index);
+
+            invoke_transfer(
+                token_prog_ai,
+                liqor_delist_token_account_ai,
+                delist_vault_ai,
+                signer_ai,
```
ckamm, line 6970: this transfer should not need the `signer_ai`, the liqor's token account authority is the liqor itself

program/src/processor.rs
```diff
@@ -6365,58 +6304,902 @@ impl Processor {
...
+            // Transfer delisting token from liqor's token account to offset liqee's borrows
+            // Transfer quote token from liqee to liqor plus fee
+
+            let max_liquidate_amount: u64 =
+                TokenAccount::load(liqor_delist_token_account_ai)?.amount.min(max_liquidate_amount);
+
+            let delist_transfer = delist_net
+                .checked_abs()
+                .unwrap()
+                .checked_ceil()
```
ckamm, line 6945: why ceil() here? At first I thought you want `delist_net` to end up between 0 and 1. But it looks like -1 to 1 is ok for dust. Then wouldn't flooring here make more sense?

ckamm (commented):
Done reading the code, but still need to think about what this all means

program/src/processor.rs
```diff
@@ -6365,56 +6303,900 @@ impl Processor {
...
+        )?;
+
+        mango_group.spot_markets[market_index].spot_market = *new_spot_market_ai.key;
+        mango_group.tokens[market_index].spot_market_mode = MarketMode::Active;
+
+        Ok(())
+    }
+
+    /// Only call this instruction if you know the sum of all quote positions is zero
+    /// There is no way to check it from program so must be manually checked
```
ckamm, line 6553: For my understanding: Would it be correct to say, that for all accounts:
- base_position must be 0 (caused by crank via force-settle)
- quote_position must be 0 (caused by crank via settle-pnl)
- funding must be 0 (done by most instructions)
- bids/asks_quantity and taker_base/quote must be 0 (because no open orders or orders on the event queue)
- mngo_accrued must be 0 (via redeem-mango)

program/src/instruction.rs
```diff
@@ -3114,6 +3114,25 @@ pub fn liquidate_delisting_token(
...
 
+pub fn create_dust_account(
+    program_id: &Pubkey,
+    mango_group_pk: &Pubkey,
+    mango_account_pk: &Pubkey,
+    payer_pk: &Pubkey,
+) -> Result<Instruction, ProgramError> {
+    let accounts = vec![
+        AccountMeta::new_readonly(*mango_group_pk, false),
+        AccountMeta::new(*mango_account_pk, false),
```
ckamm, line 3125: it's not related to this PR, but I find it suspicious that `create_dust_account()` allows creating a dust account at an arbitrary address. As far as I can tell the dust account has to be at the seed `[mango_group, "DustAccount"]`

riordanp: I believe `seed_and_create_pda` checks that the passed address matches that generated by the seeds https://github.com/blockworks-foundation/mango-v3/blob/main/program/src/processor.rs#L7544

program/src/processor.rs
riordanp: I have a script to check open interest, sum of quote/base, and MNGO accrued. Will add funding and open/uncranked orders checks to it.

program/src/processor.rs
dafyddd: Yes, it's a check we have to do before we vote on the DAO proposal. It's on the list of Sanity Checks we will do in the Delist Design doc: https://www.notion.so/mango-markets/Delist-Design-03052e7719da4a30b9b9ed7ebbbbd57f#b3ad2aff7caa4df1a7303839a0b3e37b

program/src/processor.rs
```diff
@@ -6365,58 +6304,902 @@ impl Processor {
...
+        let mut liqee_ma =
+            MangoAccount::load_mut_checked(liqee_mango_account_ai, program_id, mango_group_ai.key)?;
+        check!(!liqee_ma.is_bankrupt, MangoErrorCode::Bankrupt)?;
+        liqee_ma.check_open_orders(&mango_group, liqee_open_orders_ais)?;
+
+        // Check liqor account
+        let mut liqor_ma =
+            MangoAccount::load_mut_checked(liqor_mango_account_ai, program_id, mango_group_ai.key)?;
+        check!(
+            &liqor_ma.owner == liqor_ai.key || &liqor_ma.delegate == liqor_ai.key,
```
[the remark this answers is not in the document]
dafyddd: Yes. It's useful when you want to operate a liquidator on a machine you don't have physical possession of (e.g. a VPS). You can delegate to another private key and only put that key on the server. So if the server is compromised, at least the attacker won't be able to directly withdraw funds.

ckamm (approved):

PR #191, opened 2022-05-25
Add edit order instructions for spot and perp

dboures:
Addresses: https://trello.com/c/PsOOKgpu/130-%F0%9F%90%9E-edit-order-race-condition

This PR adds instructions that edit spot and perp order via cancelling and placing. 

To prevent multiple fills, the user sends a `cancel_order_size` which is the size that they expect to cancel. The `cancel_order_size` is compared with the remaining order size on chain and the new order will have its size reduced if the order to be cancelled has been partially filled without knowledge of the user. If the order the user is intending to cancel has been completely filled without the knowledge of the user, the edit instructions will fail.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -8,6 +8,7 @@ use anchor_lang::prelude::emit;
 use arrayref::{array_ref, array_refs};
 use bytemuck::{cast, cast_mut, cast_ref};
 use fixed::types::I80F48;
+use serum_dex::critbit::SlabView;
 use serum_dex::instruction::NewOrderInstructionV3;
 use serum_dex::state::ToAlignedBytes;
 use solana_program::account_info::AccountInfo;
@@ -2316,6 +2317,7 @@ impl Processor {
         program_id: &Pubkey,
         accounts: &[AccountInfo],
         cancel_data: Vec<u8>,
+        cancel_size: u64,
         new_order: serum_dex::instruction::NewOrderInstructionV3,
     ) -> MangoResult<()> {
         const NUM_FIXED: usize = 22;
@@ -2345,9 +2347,6 @@ impl Processor {
             msrm_or_srm_vault_ai,   // read
         ] = fixed_ais;
 
-        // TODO: make sure health + orders are correct, etc.
-
-        
         let mango_group = MangoGroup::load_checked(mango_group_ai, program_id)?;
         check_eq!(token_prog_ai.key, &spl_token::ID, MangoErrorCode::InvalidProgramId)?;
         check_eq!(dex_prog_ai.key, &mango_group.dex_program_id, MangoErrorCode::InvalidProgramId)?;
@@ -2373,7 +2372,70 @@ impl Processor {
         )?;
         let base_root_bank = RootBank::load_checked(base_root_bank_ai, program_id)?;
 
+        // load order
         let market_open_orders_ai = open_orders_ais[market_index].unwrap();
+
+        // fetch the order size, to compare it to what the user is expecting
+        let mut remaining_cancel_size: u64 = 0;
+        {
+            let open_orders = load_open_orders(market_open_orders_ai)?;
+
+            let market = load_market_state(spot_market_ai, dex_prog_ai.key)?;
+            let bids = load_bids_mut(&market, bids_ai)?;
+            let asks = load_asks_mut(&market, asks_ai)?;
+
+            for j in 0..128 {
+                let slot_mask = 1u128 << j;
+                if open_orders.free_slot_bits & slot_mask != 0 {
+                    // means slot is free
+                    continue;
+                }
+                let order_id = open_orders.orders[j];
+
+                if open_orders.is_bid_bits & slot_mask != 0 {
+                    match bids
+                        .find_by_key(order_id)
+                        .and_then(|order| bids.get(order))
+                        .and_then(|node| node.as_leaf())
+                    {
+                        None => continue,
+                        Some(bid) => {
+                            msg!("bid: {:?}", bid);
+                            remaining_cancel_size = bid.quantity();
+                            break;
+                        }
+                    };
+                } else {
+                    match asks
+                        .find_by_key(order_id)
+                        .and_then(|order| asks.get(order))
+                        .and_then(|node| node.as_leaf())
+                    {
+                        None => continue,
+                        Some(ask) => {
+                            msg!("ask: {:?}", ask);
+                            remaining_cancel_size = ask.quantity();
+                            break;
+                        }
+                    };
+                };
+            }
+        }
+
+        msg!("remaining_cancel_size: {:?}", remaining_cancel_size);
+        msg!("cancel_size: {:?}", cancel_size);
+
+        // cancel order (if size matches)
+
+        // settle funds?
+
+        // place order pre checks
+
+        // place order
+
+        // update health, etc
+
+        // TODO: make sure health + orders are correct, etc.
 
         let order_side = new_order.side;
         let vault_ai = match order_side {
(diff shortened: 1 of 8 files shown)


microwavedcola1 (commented):
I did a first pass, will do another in 1-2 days. Then we could approve

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+
+            for j in 0..128 {
+                let slot_mask = 1u128 << j;
+                if open_orders.free_slot_bits & slot_mask != 0 {
+                    // means slot is free
+                    continue;
+                }
+                let order_id = open_orders.orders[j];
+
+                if open_orders.is_bid_bits & slot_mask != 0 {
```
ckamm, line 2378: why even call `find_by_key()` for orders that don't have the same side as `new_order.side` or `cancel_order.side`?

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+            let bids = load_bids_mut(&market, bids_ai)?;
+            let asks = load_asks_mut(&market, asks_ai)?;
+
+            for j in 0..128 {
+                let slot_mask = 1u128 << j;
+                if open_orders.free_slot_bits & slot_mask != 0 {
+                    // means slot is free
+                    continue;
+                }
+                let order_id = open_orders.orders[j];
```
ckamm, line 2376: why are we looking at orders that have `order_id != cancel_order.order_id`? Wouldn't this find the quantity of some random other open order?

maybe we can fully skip iterating over the 128 potential open orders and check bids/asks directly?

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+                };
+            }
+        }
+
+        check!(expected_cancel_size > 0, MangoErrorCode::Default)?;
+        check!(remaining_cancel_size > 0, MangoErrorCode::Default)?;
+        check!(new_order.side == cancel_order.side, MangoErrorCode::Default)?;
+
+        // if cancel order has been partially filled, reduce the new order size to accommodate
+        let filled_amount = expected_cancel_size.checked_sub(remaining_cancel_size).unwrap();
```
ckamm, line 2411: minor: possibly a non-panic error when expected < remaining

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+            new_order.max_coin_qty = NonZeroU64::new(new_max_qty).unwrap();
+            match cancel_order.side {
+                serum_dex::matching::Side::Bid => {
+                    let new_max_pc_qty_including_fees = pc_lot_size
+                        .checked_mul(new_order.limit_price.get())
+                        .unwrap()
+                        .checked_mul(new_order.max_coin_qty.get())
+                        .unwrap();
+                    new_order.max_native_pc_qty_including_fees =
+                        NonZeroU64::new(new_max_pc_qty_including_fees).unwrap();
```
ckamm, line 2428: This just overrides whatever the user set as max-coin-qty, and doesn't account for any fees: seems suspicious but I'll need to double-check serum

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+                }
+                serum_dex::matching::Side::Ask => {
+                    new_order.max_native_pc_qty_including_fees =
```
ckamm, line 2431: if there's ever any use to setting this to something else than `u64::max`, we're overriding the users choice here.

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+                throw_err!(MangoErrorCode::Default);
+            }
+
+            let filled_amount = expected_cancel_size.checked_sub(book_order.quantity).unwrap();
+            if filled_amount > 0 {
+                let new_max_base = max_base_quantity.checked_sub(filled_amount).unwrap();
+                if new_max_base < 0 {
+                    throw_err!(MangoErrorCode::Default);
+                }
+                max_base_quantity = new_max_base;
```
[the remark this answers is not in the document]
ckamm: definitely, max_base_quantity could be i64::MAX, with max_quote_quantity being used for the order limiting

program/src/processor.rs
```diff
@@ -2311,6 +2313,385 @@ impl Processor {
...
+
+            let market_index = mango_group.find_perp_market_index(perp_market_ai.key).unwrap();
+
+            let book = Book::load_checked(program_id, bids_ai, asks_ai, &perp_market)?;
+
+            let (order_id, cancel_side) = mango_account
+                .find_order_with_client_id(market_index, client_order_id)
+                .ok_or(throw_err!(MangoErrorCode::ClientIdNotFound))?;
+
+            let book_order = book.get_order(order_id, side)?;
```
ckamm, line 2539: this needs to deal with expired orders: maybe the old order is still on the book, but expired?

program/src/processor.rs
```diff
@@ -6832,6 +7213,96 @@ impl Processor {
...
+                    max_quote_quantity,
+                    expiry_timestamp,
+                    client_order_id,
+                    cancel_order_size,
+                    side,
+                    order_type,
+                    reduce_only,
+                    limit,
+                );
+                if invalid_id_ok {
```
ckamm, line 7254: hmm. if the previous order is missing, should we assume it was fully filled and adjust the new order accordingly?

Current behavior seems to be:
- if the old order is partially filled, cancel it; adjust new order and place it
- if the old order is missing:
  - if invalid_id_ok is false: error
  - if invalid_id_ok is true: do nothing (i.e. don't place the new order)

To me it sounds like "if the old order is gone (fully filled), then adjust the new order for that and place it" would also make sense.

cloorem2:
this PR looks useful. i'm seeing plenty of double fills.

program/src/processor.rs
dboures: Great suggestion, I have updated the logic to look directly at the bids/asks

program/src/processor.rs
dboures: Initially I copied the invalid_id_ok flag to mimic the cancel_perp_order instructions, but now I think it doesn't make much sense for editing orders. To be more straightforward with the user, edit_order instructions should fail if there is no order to edit. The updated behavior is:

- if the old order is partially filled, cancel it; adjust new order and place it
- if the old order is missing (fully filled), instruction will fail because ```book.get_order()``` will fail to find an order and will throw a ```MangoErrorCode::InvalidOrderId```

program/src/processor.rs
dboures: I thought it's ok to edit/update an expired order as long as the new order is not already expired (and ```place_perp_order2``` checks for this)? But let me know if this is not the case. I'm not sure what the behavior should be if the order is expired (maybe just throw an error?)

program/src/processor.rs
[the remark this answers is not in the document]
dboures: I'm not sure what is best here. Initially, I tried to follow the example from the client: https://github.com/blockworks-foundation/mango-client-v3/blob/93240e04c6c87465758a9e02e1637d2f487d05aa/src/client.ts#L1961, where  maxQuoteQuantity is optional, and defaults to i64::MAX. We could perhaps check if ```max_quote_quantity != i64::MAX``` and then adjust downwards as necessary, but I wasn't sure at which price I should convert filled_amount into quote? Unless I have it backwards as currently written?

program/src/processor.rs
dboures: This and the below comment have been updated to properly include fees and not overwrite user inputs

NostalgicGareth (approved):

PR #193, opened 2022-05-31
PlacePerpOrder2: Add flag to request relative expiry

ckamm:
This makes it easier to use TIF when the solana cluster clock does not match the actual clock.

It's not a perfect substitute, since relative expiry will be relative to transaction execution and not to transaction sending.


microwavedcola1 (approved):
1/ lgtm, we could adapt the place_perp_order2 ix helper..
2/ did the cargo audit always fail?

ckamm:
> 1/ lgtm, we could adapt the place_perp_order2 ix helper.. 2/ did the cargo audit always fail?

What did you mean about the ix helper?

PR #195, opened 2022-06-02
add Withdraw2 instruction using compact OO layout

mschneider:


ckamm (commented):
Looks good, but I'll need to compare withdraw() and withdraw2() in more detail

ckamm:
I wish these withdraw functions could share code, but that's harder to do than it looks.

PR #201, opened 2022-08-11
Fix remove_oracle market state checks

riordanp:
remove_oracle fails for spot only markets since their market state is always Default. This PR changes the check to look at market pubkey instead, which should be the default if the market has either never been added, or has been removed with remove_spot/perp_market


ckamm (approved):
This looks correct to me.

Both `remove_spot_market` and `remove_prep_market` set the pubkey to `Pubkey::default()` and set the market_mode to Inactive. The problem was that never-active markets had `market_mode == Default`.

PR #202, opened 2022-08-19
Add second fee tier

riordanp:
- Adds an optional second taker fee tier
- Adds a new instruction `change_referral_fee_params2` to set the second fee tier parameters

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -6211,7 +6211,7 @@ impl Processor {
         let mut mango_group = MangoGroup::load_mut_checked(mango_group_ai, program_id)?;
         check_eq!(admin_ai.key, &mango_group.admin, MangoErrorCode::InvalidAdminKey)?;
         check!(admin_ai.is_signer, MangoErrorCode::SignerNecessary)?;
-        msg!("old referral fee params: ref_surcharge_centibps_tier_1: {} ref_share_centibps_tier_1: {} ref_surcharge_centibps_tier_2: {} ref_share_centibps_tier_2: {} ref_mngo_required: {} ref_mngo_tier_2_factor {}", mango_group.ref_surcharge_centibps_tier_1, mango_group.ref_share_centibps_tier_1, mango_group.ref_surcharge_centibps_tier_2, mango_group.ref_share_centibps_tier_2, mango_group.ref_mngo_required, mango_group.ref_mngo_required_tier_2_factor);
+        msg!("old referral fee params: ref_surcharge_centibps_tier_1: {} ref_share_centibps_tier_1: {} ref_surcharge_centibps_tier_2: {} ref_share_centibps_tier_2: {} ref_mngo_required: {} ref_mngo_tier_2_factor {}", mango_group.ref_surcharge_centibps_tier_1, mango_group.ref_share_centibps_tier_1, mango_group.ref_surcharge_centibps_tier_2, mango_group.ref_share_centibps_tier_2, mango_group.ref_mngo_required, mango_group.ref_mngo_tier_2_factor);
 
         // TODO - when this goes out, if there are any events on the EventQueue fee logging will be messed up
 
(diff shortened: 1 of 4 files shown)


program/src/processor.rs
```diff
@@ -6146,18 +6146,67 @@ impl Processor {
...
+    ) -> MangoResult {
+        
+        let tier_2_enabled = ref_surcharge_centibps_tier_2 != 0 && ref_share_centibps_tier_2 != 0;
+        if tier_2_enabled {
+            // tier 1 surcharge must be less than tier 2 surcharge so 10k holders get a discount
+            check!(ref_surcharge_centibps_tier_1 < ref_surcharge_centibps_tier_2, MangoErrorCode::InvalidParam)?;
+            // tier 1 share must be greater than tier 1 surcharge so 10k referees pay more then their referrer
+            check!(ref_share_centibps_tier_1 > ref_surcharge_centibps_tier_1, MangoErrorCode::InvalidParam)?;
+            // tier 2 share must be less than tier 2 surcharge so 100k referees don't pay more than base fee
+            check!(ref_share_centibps_tier_2 < ref_surcharge_centibps_tier_2, MangoErrorCode::InvalidParam)?;
```
ckamm, line 6180: The lack of symmetry to the above condition is surprising!

Explained to myself like this: share_1 and surcharge_1 are for the 10k MNGO level. But share_2 is for referrers that have more than 100k MNGO while surchare_2 is for accounts that have nothing.

Add a condition that share_2 >= share_1?

Don't we want a total ordering where surcharge_1 <= share_1 <= share_2 <= surcharge_2?

program/src/state.rs
```diff
@@ -51,6 +51,7 @@ pub const MAX_NUM_IN_MARGIN_BASKET: u8 = 9;
 pub const INDEX_START: I80F48 = I80F48!(1_000_000);
 pub const PYTH_CONF_FILTER: I80F48 = I80F48!(0.10); // filter out pyth prices with conf > 10% of price
 pub const CENTIBPS_PER_UNIT: I80F48 = I80F48!(1_000_000);
+pub const REF_TIER_2_FACTOR: I80F48 = I80F48!(10);
```
ckamm, line 54: sad that this needs to be hardcoded, I guess making the share_2/surcharge_2 u16 is too limiting...

program/src/state.rs
```diff
@@ -1198,8 +1199,16 @@ impl HealthCache {
                 let mngo_cache = &mango_cache.root_bank_cache[mngo_index];
                 let mngo_deposits = mango_account.get_native_deposit(mngo_cache, mngo_index)?;
                 let ref_mngo_req = I80F48::from_num(mango_group.ref_mngo_required);
-                if mngo_deposits < ref_mngo_req {
-                    market_fees += (I80F48::from_num(mango_group.ref_surcharge_centibps)
+                let tier_2_enabled = mango_group.ref_surcharge_centibps_tier_2 != 0 && mango_group.ref_share_centibps_tier_2 != 0;
+
+                if mngo_deposits < ref_mngo_req * REF_TIER_2_FACTOR {
+                    let surcharge = if tier_2_enabled && mngo_deposits > ref_mngo_req {
```
ckamm, line 1205: I think this should be `mngo_deposits < ref_mngo_req`

ckamm (changes requested):

program/src/state.rs
riordanp: Yes it's not ideal. I actually think u16 would be fine for the values that would realistically be used, just went with 32 for consistency. Wonder if I can change this now without breaking devnet...

program/src/processor.rs
riordanp: Added an extra condition

ckamm (approved):

program/src/state.rs
ckamm: could set to 0 on devnet, change the program, re-set

riordanp: Tried this out, but unfortunately the MNGO required in native units will overflow a u32 as well. Since we have the quote decimals already on the group and the threshold is likely to be a large round number, we could store the 'UI' value and convert to native where required. It's a bit messy since tier 1 and tier 2 would then be in different units so I'm not sure I like this solution.

riordanp: Trying this out today, switching tier 2 params to u16 and storing the factor on the group as a u8

PR #208, opened 2022-10-12
cli script to compute account equity from solana snapshot

ckamm:
don't merge, review only


cli/src/main.rs
```diff
@@ -0,0 +1,529 @@
...
+            let (owner, equity) = equity_opt.unwrap();
+            account_equities.push((account_address, owner, equity));
+        }
+
+        let token_names: [&str; 16] = [
+            "MNGO", "BTC", "ETH", "SOL", "USDT", "SRM", "RAY", "COPE", "FTT", "ADA", "MSOL", "BNB",
+            "AVAX", "LUNA", "GMT", "USDC",
+        ];
+
+        let available_tokens: [bool; 15] = [
```
microwavedcola1, line 175: why is this an array of size 15 and not 16?

cli/src/main.rs
```diff
@@ -0,0 +1,529 @@
...
+            true, true, false, // cope delisted
+            true, false, // no spot ada
+            true, true, true, false, // luna delisted
+            true,
+        ];
+
+        //println!("{:?}", ctx.cache.price_cache.iter().map(|c| c.price).collect::<Vec<_>>());
+
+        // TODO: tentative numbers from "Repay bad Debt #2" proposal
+        let available_native_amounts: [u64; 15] = [
```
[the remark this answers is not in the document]
ckamm: Index 15 is USDC, where we don't have a specific target number of tokens available. Instead the amount needed for full coverage is an output of the script.

cli/src/main.rs
```diff
@@ -0,0 +1,761 @@
...
+            60800000000,
+            180900000000,
+            0,
+            152843000000000,
+        ];
+
+        // Token prices at time of reimbursement
+        // Note that user equity at snapshot time is computed from the prices from the
+        // mango cache in the snapshot.
+        let reimbursement_prices: [I80F48; 16] = [
```
dsemerson, line 293: Are these values going to be hardcoded right before the script gets kicked off? If so I presume multiple people would be verifying the values are correct? Would it be better to reference an oracle at runtime? 

I am a little bit scared that the script is relying on matching up indexes across multiple hardcoded arrays. They do seem to be correctly matched up though, and I suppose the reimbursement prices can have some margin of error.

ckamm: Yes, the prices would be frozen and updated as close to actual reimbursement as we can. I don't think pulling from an oracle is a good idea.

Constants: I agree and had meant to clean that up for a while. Done now.

kronael:
Checked our balances for https://docs.google.com/spreadsheets/d/131YQnLUBgatfhqfpf1FQfuRE8SNpdLq1sqraOAyKfLM/edit#gid=1057746867 ... seems legit.  Thanks.

PR #209, opened 2022-10-17
Allow recovery of funds from mainnet.1 group

soapyp:
- Add RecoveryForceSettleSpotOrders instruction
- Add RecoveryWithdrawTokenVault instruction
- Add RecoveryWithdrawMngoVault instruction
- Add RecoveryWithdrawInsuranceVault instruction
- Re-enable WithdrawMsrm

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -7566,7 +7566,7 @@ impl Processor {
         let accounts = array_ref![accounts, 0, NUM_FIXED];
         let [
             mango_group_ai,     // read
-            token_source_ai,    // write
+            vault_ai,           // write
             token_account_ai,   // write
             root_bank_ai,       // read
             node_bank_ai,       // read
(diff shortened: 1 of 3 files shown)


program/src/processor.rs
```diff
@@ -7412,11 +7412,302 @@ impl Processor {
...
+            (
+                open_orders.native_coin_free,
+                open_orders.native_pc_free + open_orders.referrer_rebates_accrued,
+            )
+        };
+
+        if pre_base == 0 && pre_quote == 0 {
+            // margin basket may be in an invalid state; correct it before returning
+            let open_orders = load_open_orders(open_orders_ai)?;
+            mango_account.update_basket(market_index, &open_orders)?;
```
mschneider, line 7515: should emit a log message as well, so the crank knows not to retry this account

